PR #40, opened 2023-02-26
Requirements

joeleu:
Apologies for the large request size, I promise to not make a habit out of that. This presents a working registration flow for both the Organization and Volunteer user types along with a login page and profile page that loads based on type.


Hgersten-hash (commented):
It looks good to me. I would add a few more comments for readability, and now that you added the validations for the form I think it is ready.

PR #53, opened 2023-02-28
replaced key with safer option

JinweiZhu:


voluncheer/requirements.txt
```diff
@@ -4,4 +4,6 @@ django_debug_toolbar==3.8.1
 Django==3.2.8
 pytz==2022.7.1
 sqlparse==0.4.3
-pysqlite3-binary
+#Not installable by pip
+#pysqlite3-binary
```
joeleu, line 8: I was able to install this via pip:

pysqlite3-binary==0.5.0

joeleu (commented):
Have you considered using the builtin GitHub secret management system?

Ref: https://docs.github.com/en/codespaces/managing-your-codespaces/managing-encrypted-secrets-for-your-codespaces

voluncheer/requirements.txt
JinweiZhu: I will put it back in my later branch pull request.

JinweiZhu:
> Have you considered using the builtin GitHub secret management system?
> 
> Ref: https://docs.github.com/en/codespaces/managing-your-codespaces/managing-encrypted-secrets-for-your-codespaces

Are you familiar with Github Action and Github workflow? That seems to be the only way to bring in the secret key as an environment variable.

voluncheer/requirements.txt
joeleu: Actually in light of moving toward PostgresSQL you can just delete it.

joeleu:
For posterity, we discussed this during the scrum and we decided to pass around the secret that we will all store locally.

voluncheer/requirements.txt
JinweiZhu: Will remove in the future when that happens.

joeleu: What purpose does it serve as a comment? There is no sense in leaving it commented out. If not having it is an issue we add it back.

JinweiZhu: Sorry, typo. Fixed.

PR #56, opened 2023-03-03
Eden: WIP edit profile

EdenWuyifan:


voluncheer/profiles/forms/organizations.py
```diff
@@ -1,5 +1,5 @@
 from django import forms
-from django.contrib.auth.forms import UserCreationForm
+from django.contrib.auth.forms import UserCreationForm, UserChangeForm
```
joeleu, line 2: Can you put these on separate lines? It makes future refactoring easier.

voluncheer/profiles/forms/organizations.py
```diff
@@ -25,3 +25,16 @@ def save(self, commit=True):
             name=self.cleaned_data.get("name"),
         )
         return user
+
+class OrganizationChangeForm(UserChangeForm):
```
joeleu, line 29: Please add a docstring, even if it's a simple one-liner. Here and elsewhere.

https://peps.python.org/pep-0008/#documentation-strings

voluncheer/profiles/urls.py
```diff
@@ -2,15 +2,16 @@
 from django.urls import path
 
 from profiles.views.home import home
-from profiles.views.profile import ProfileView
+from profiles.views.profile import *
 
 urlpatterns = [
     path("", home, name="home"),
-    path("profile/", ProfileView.as_view(), name="profile"),
+    path(r'^profile/$', ProfileView.as_view(), name="profile"),
```
joeleu, line 9: Please keep the quote we are using the same unless it would require you to escape a quote within a string. We have been using the double-quote. Here and elsewhere.

joeleu, line 9: Do we really need these anchors?

voluncheer/profiles/urls.py
```diff
@@ -2,15 +2,16 @@
 from django.urls import path
 
 from profiles.views.home import home
-from profiles.views.profile import ProfileView
+from profiles.views.profile import *
```
joeleu, line 5: Please enumerate the actual imports rather than using a wild card. It makes identifying what is used where much easier.

voluncheer/profiles/views/profile.py
```diff
@@ -1,11 +1,14 @@
 from django.contrib.auth.decorators import login_required
```
joeleu, line 1: Please keep imports sorted alphabetically. The exception is to keep external imports (one's we are not writing) separate from internal imports (the modules we are writing).

voluncheer/profiles/views/profile.py
```diff
@@ -34,4 +38,15 @@ def get_context_data(self, **kwargs):
...
+            kwargs['user_form'] = VolunteerChangeForm(instance=self.request.user)
         return super().get_context_data(**kwargs)
+
+
+def form_valid(request):
+    """Saves the new user and logs them in."""
+    if request.user.is_volunteer:
+        form = VolunteerChangeForm(request.POST, instance=request.user)
+    elif request.user.is_organization:
+        form = OrganizationChangeForm(request.POST, instance=request.user)
```
joeleu, line 50: We should include the `else` block to catch users that do not match those types. e.g. admin.

joeleu (changes requested):

voluncheer/profiles/forms/organizations.py
EdenWuyifan: Sure thing! Thanks for the correction

voluncheer/chatroom/admin.py
```diff
@@ -1,3 +1,3 @@
-from django.contrib import admin
+# from django.contrib import admin
```
joeleu, line 1: Instead of commenting these out we should remove these files until we need them. There is no sense in having modules that don't do anything.

Here and elsewhere.

voluncheer/chatroom/urls.py
```diff
@@ -2,9 +2,17 @@
 
 from . import views
 
-app_name = 'chatroom'
+app_name = "chatroom"
```
joeleu, line 5: Does this actually do anything? I can't find where it is used. Plus it doesn't match Python's style guide for constants which is UPPER_SNAKE_CASE.

voluncheer/job_board/models.py
```diff
@@ -2,9 +2,11 @@
 
 
 class Job(models.Model):
-    organization = models.ForeignKey("profiles.Organization", on_delete=models.CASCADE)
+    organization = models.ForeignKey(
+        "profiles.Organization", on_delete=models.CASCADE
+    )  # noqa: E501
```
joeleu, line 7: We should really try hard to avoid adding in exceptions to the linter. There is almost always a way to make it work. If you need help figuring out how to make it fit the style preference just let me know and we can figure it out.

Here and elsewhere.

voluncheer/profiles/forms/organizations.py
```diff
@@ -26,8 +27,16 @@ def save(self, commit=True):
         )
         return user
 
+
+""" OrganizationChangeForm
```
joeleu, line 31: This should go just below the class definition rather than above it. The first line should also give a brief description of the class, the name is obvious from the definition.

Here and elsewhere.

voluncheer/profiles/models.py
```diff
@@ -5,13 +5,15 @@
 
 class UserType(models.IntegerChoices):
     """Container for all possible user types."""
+
```
joeleu, line 8: Is this suggested by the linter? I usually try to avoid unnecessary white space.

voluncheer/profiles/urls.py
```diff
@@ -2,16 +2,35 @@
...
-        # Here we can place organization only URLs.
-    ], "profiles"), namespace="organizations")),
-    path("volunteers/", include(([
-        # Here we can place volunteer only URLs.
-    ], "profiles"), namespace="volunteers")),
-    path(r'^profile/update/$', form_valid, name="profile_update"),
-]
\ No newline at end of file
+    path(r"^profile/$", ProfileView.as_view(), name="profile"),
+    path(
```
joeleu, line 11: We can actually just remove this since it isn't doing anything. I mistakenly left it in.

voluncheer/voluncheer/settings.py
```diff
@@ -139,8 +141,11 @@
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/4.1/howto/static-files/
-STATIC_ROOT = os.path.join(BASE_DIR, "static")
-STATIC_URL = '/static/'
+# STATIC_ROOT = os.path.join(BASE_DIR, "static")
```
joeleu, line 144: Is this unused? If so it should be removed not commented out.

joeleu (changes requested):

voluncheer/job_board/views.py
```diff
@@ -3,61 +3,18 @@
 
 from job_board.models import Job
 
-Organization = apps.get_model('profiles', 'Organization')
+Organization = apps.get_model("profiles", "Organization")
 
 # ======================== Job Board ============================
-test_jobs = [
```
JinweiZhu, line 9: Why are we removing these commented out tests? Are you sure we won't use it later?

voluncheer/job_board/views.py
```diff
@@ -3,61 +3,18 @@
 
 from job_board.models import Job
```
JinweiZhu, line 4: I think job_board folder has become jobboard folder. The original job_board should be removed, right? It's no longer in develop branch.

JinweiZhu (commented):
Good job.

voluncheer/chatroom/urls.py
EdenWuyifan: This is done by Flask8 autocorrect on vs code, so I suppose this is what flask8 told us to follow?

voluncheer/job_board/models.py
EdenWuyifan: Sure thing, the wrap is also done by autocorrect on vs code. The # noqa: E501 I added is for flask8 error ignoration.

voluncheer/profiles/models.py
EdenWuyifan: Again, flask8 autocorrect

voluncheer/profiles/urls.py
EdenWuyifan: Sure

voluncheer/voluncheer/settings.py
EdenWuyifan: Thanks for noticing, my mistake, I commented that only for local tests

voluncheer/job_board/views.py
EdenWuyifan: Yes, I suppose we need to change our job models in the future, and it causes warnings in flask8

voluncheer/job_board/views.py
EdenWuyifan: Not on develop branch by my observation, are you sure you merged it?

PR #57, opened 2023-03-05
"Added tests for profile models, forms and views"

dbleek:
I split Joe's initial tests into three files for models, forms and views, adding the following:

Models:
Test that date of birth field is filled properly
Test that overridden createsuperuser() works
Test that model object name functions return correct value
Test that email max length = 254
Test that user cannot be created without email address

Forms:
Test that all volunteer/organization forms are properly labeled
Test that regex validator for volunteer name is working
Test that new users must submit a unique email address to register

Views:
Test that the signup, volunteer signup, and organization signup views are all accessible and use the correct template
Test that users are redirected to login if not authenticated


voluncheer/profiles/test_forms.py
```diff
@@ -0,0 +1,108 @@
+from django.test import TestCase
+
+from profiles.forms.volunteers import VolunteerCreationForm
+from profiles.forms.organizations import OrganizationCreationForm
+
+
+class OrganizationCreationFormTest(TestCase):
```
joeleu, line 7: Please add class level docstrings: https://peps.python.org/pep-0008/#documentation-strings

Here and for the other classes. They can be basic one-liners.

voluncheer/profiles/test_forms.py
```diff
@@ -0,0 +1,108 @@
...
+
+class VolunteerCreationFormTest(TestCase):
+    def setUp(self):
+        """See base class."""
+        self.form = VolunteerCreationForm()
+
+    def test_first_name_label(self):
+        """Test first name field is labeled correctly"""
+        self.assertTrue(
+            self.form.fields["first_name"].label is None
```
joeleu, line 51: Why do we need to check if the label is `None` OR the expected value?

voluncheer/profiles/test_models.py
```diff
@@ -80,4 +84,43 @@ def test_organization_details(self):
...
+        """Tests admin account has proper attributes."""
+        self.assertTrue(self.admin.is_staff)
+        self.assertTrue(self.admin.is_superuser)
+        self.assertEqual(self.admin.type, UserType.ADMIN)
+
+    def test_email_is_required(self):
+        """Tests creating a user without an email raises desired ValueError"""
+        with self.assertRaisesMessage(
+            ValueError, "creating a user requires an email field"
+        ):
```
joeleu, line 101: The actual error message here is not particularly important, I'd prefer just asserting the correct error type is raised. Specifying the actual content is only really useful when it is dynamic. When the content is static the test serves as a change detector which is an anti-pattern.

joeleu (changes requested):

voluncheer/profiles/test_forms.py
```diff
@@ -0,0 +1,108 @@
...
+
+class OrganizationCreationFormTest(TestCase):
+    def setUp(self):
+        """See base class."""
+        self.form = OrganizationCreationForm()
+
+    def test_name_label(self):
+        """Test name field is labeled correctly"""
+        self.assertTrue(
+            self.form.fields["name"].label is None
```
JinweiZhu, line 15: same as Joe's comment below.

voluncheer/profiles/test_models.py
```diff
@@ -1,16 +1,12 @@
...
 from profiles.models import User
 from profiles.models import UserType
 from profiles.models import Volunteer
 
-from map import urls
-from map import views
 
-'''
+"""
 Commented out for now until I figure out the namespace and app_name diff
```
JinweiZhu, line 10: If you are already testing for the view below, then these tests for URLs might not be necessary. Pls double-check and remove if that's the case.

voluncheer/profiles/test_views.py
```diff
@@ -0,0 +1,67 @@
+from django.test import TestCase
+from django.urls import reverse
+
+from profiles.models import Organization
+from profiles.models import User
+from profiles.models import UserType
+from profiles.models import Volunteer
+
+
+class SignupViewTest(TestCase):
```
JinweiZhu, line 10: Would you like to move the url from voluncheer to profiles as well, since these tests are built in voluncheer?

voluncheer/profiles/test_forms.py
dbleek: According to the tutorial I followed:

"Note here that we also have to test whether the label value is None, because even though Django will render the correct label it returns None if the value is not explicitly set."

I agree it's a bit strange to test for the default None value since the labels are not explicitly set in any of the models, but I suppose it protects us if we someday decide to add labels and they are entered incorrectly. Also happy to remove the tests if we think they are just fluff.

voluncheer/profiles/test_forms.py
dbleek: See above.

voluncheer/profiles/test_forms.py
dbleek: Added.

voluncheer/profiles/test_models.py
dbleek: Yes, thank you for pointing out. Removed.

voluncheer/profiles/test_views.py
dbleek: We can change in a follow-up pull request, will add an issue.

voluncheer/profiles/test_models.py
dbleek: Thank you for pointing out, changed to assertRaises().

dbleek:
I removed the test that confirmed that attempting to go a profile when not logged in goes to login for now. It appears the profile urls have been changed and would like to discuss with Eden before trying to fix.

voluncheer/profiles/test_views.py
JinweiZhu: Sounds good.

voluncheer/profiles/test_forms.py
joeleu: I don't like it but if that's what Django recommends we should stick with it. Thanks for the explanation.

joeleu (approved):

PR #80, opened 2023-03-12
Separate CI and CD

joeleu:
This PR separates out the Continuous Integration testing from the Continuous Deployment GitHub Actions Workflows. Along with separation the trigger for deploying to AWS is changed from `pull_request` to `push` to avoid missing merges that occur after a pull request was executed on a different branch. Finally, this bumps the coverage minimum up to the required minimum (85%) and removes some dead, unused decorators. I added these during the first commit but we haven't had a need for them so they are just taking up space and cognitive overhead so they are being removed.


.github/actions/test/action.yml
```diff
@@ -1,5 +1,10 @@
 name: Test Python
 description: Runs Python tests and checks coverage when tests pass
+inputs:
```
dbleek, line 3: Shouldn't this be required? Or is it not because of the default?

.github/workflows/presubmit.yml
```diff
@@ -25,13 +16,9 @@ concurrency:
...
   cancel-in-progress: true
 
-permissions:
-  id-token: write
-  contents: read
-
 jobs:
-  format-lint-test:
-    name: Checks format, lints, runs tests, and checks coverage
+  format-lint-and-test:
```
dbleek, line 20: Not that it really matters, but are we linting PRs twice then? Once on submission, once again when pushing to develop or main?

.github/actions/test/action.yml
```diff
@@ -12,7 +17,7 @@ runs:
       working-directory: voluncheer
       run: |
         coverage run --source "." manage.py test
-        coverage lcov --fail-under=${MINIMUM_COVERAGE}
+        coverage lcov --fail-under=${{ inputs.minimum-coverage }}
```
dbleek, line 20: Should this be INPUTS_MINIMUM-COVERAGE? See: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs

dbleek (approved):
This looks good as far as I can tell.  Did a bit of dive into the Github Actions documentation, I commend you on picking it up so quickly.

.github/actions/test/action.yml
joeleu: I've not marked it required because there is a default.

.github/actions/test/action.yml
joeleu: Maybe both are supported?

This is the reference I was using with regards to these contexts: https://docs.github.com/en/actions/learn-github-actions/contexts#inputs-context

.github/workflows/presubmit.yml
joeleu: That's right. The risk I'm trying to mitigate with that is folks pushing directly to the develop and/or main branches. Until we can get branch protection rules in place this remains a possibility. I know this because I screwed up and pushed directly to the develop branch /facepalm

PR #82, opened 2023-03-13
Post an Opportunity

dbleek:
Changed all mentions of job to opportunity, along with any other changes from Eden's post a job commit.


voluncheer/.gitignore
```diff
@@ -0,0 +1,9 @@
+.idea/
```
joeleu, line 1: This file is redundant with the .gitignore in the root of the repo and can be removed.

voluncheer/opportunityboard/forms/postanopportunity.py
```diff
@@ -0,0 +1,57 @@
+from django import forms
+from django.contrib.auth.decorators import login_required
+from django.utils import timezone
+
+from profiles.models import Organization
```
joeleu, line 5:
```suggestion
from opportunityboard.models import Opportunity
from profiles.models import Organization
```

voluncheer/opportunityboard/forms/postanopportunity.py
```diff
@@ -0,0 +1,57 @@
...
+
+from opportunityboard.models import Opportunity
+
+
+class PostAnOpportunityForm(forms.ModelForm):
+    """This is the form used for creating a new opportunity for organization users"""
```
joeleu, line 11: Please add punctuation to the docstring.

voluncheer/opportunityboard/forms/postanopportunity.py
```diff
@@ -0,0 +1,57 @@
...
+        ),
+    )
+
+    class Meta:
+        model = Opportunity
+        fields = (
+            "category",
+            "title",
+            "description",
+            # "date",
```
joeleu, line 35: date and duration should either be used or deleted, not commented out.

voluncheer/opportunityboard/forms/postanopportunity.py
```diff
@@ -0,0 +1,57 @@
...
+        user = self.instance
+        organization = Organization.objects.get(pk=user)
+        if self.is_valid():
+            opportunity = Opportunity()
+            opportunity.pubdate = timezone.now()
+            opportunity.organization = organization
+            opportunity.category = self.cleaned_data.get("category")
+            opportunity.title = self.cleaned_data.get("title")
+            opportunity.description = self.cleaned_data.get("description")
+            # opportunity.date = self.cleaned_data.get("date")
```
joeleu, line 52: Same as above.

voluncheer/opportunityboard/test_models.py
```diff
@@ -3,13 +3,13 @@
 from profiles.models import Organization
 from profiles.models import User
 from profiles.models import UserType
-from jobboard.models import Job
+from opportunityboard.models import Opportunity
```
joeleu, line 6: Keep sorted alphabetically. I'm really surprised our linter doesn't require this.

```suggestion
from opportunityboard.models import Opportunity
from profiles.models import Organization
from profiles.models import User
from profiles.models import UserType
```

voluncheer/profiles/test_forms.py
```diff
@@ -1,7 +1,23 @@
+from datetime import datetime
+
+from datetime import datetime
+
 from django.test import TestCase
 
+from profiles.models import Organization
```
joeleu, line 7: Please sort alphabetically.

voluncheer/requirements.txt
```diff
@@ -17,6 +17,7 @@ django-widget-tweaks==1.4.12
 django_debug_toolbar==3.8.1
 flake8==5.0.4
 jmespath==1.0.1
+mock==5.0.1
```
joeleu, line 20: I don't think mock is used and thus should be removed. Plus, mock is now built into the `unittest` package: https://docs.python.org/3/library/unittest.mock.html

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -0,0 +1,20 @@
+from django.shortcuts import render
+from django.apps import apps
+
+from opportunityboard.models import Opportunity
+
+Organization = apps.get_model("profiles", "Organization")
+
+# ======================== Opportunity Board ============================
```
joeleu, line 8: This comment doesn't really add anything, please remove.

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -0,0 +1,20 @@
+from django.shortcuts import render
```
joeleu, line 1: Please sort alphabetically.

voluncheer/opportunityboard/views/postajob.py
```diff
@@ -0,0 +1,22 @@
+from django.http import Http404
+from django.shortcuts import render, redirect
+
+from opportunityboard.forms.postajob import PostAnOpportunityForm
+
+
+def post_a_job(request):
```
joeleu, line 7: I think this file was meant to be deleted.

voluncheer/opportunityboard/urls.py
```diff
@@ -0,0 +1,13 @@
...
+from opportunityboard.views.opportunityboard import opportunityboard, select
+from opportunityboard.views.postanopportunity import post_an_opportunity
+
+
+urlpatterns = [
+    path("", opportunityboard, name="opportunityboard"),
+    # Opportunityboard
+    path("select", select, name="select"),
+    # Post an Opportunity
+    path("post_an_opportunity", post_an_opportunity, name="post_an_opportunity"),
```
joeleu, line 12: So this is certainly the thing I understand least about django. Is this path (post_an_opportunity) the URL path the user will see/enter in the browser? If so, we should rethink it but if it is just the path the app directs to internally it is fine.

joeleu (changes requested):

voluncheer/opportunityboard/urls.py
EdenWuyifan: Yes, it is an url path. For now we can add guard clauses to our views in order to defend unauthorized users. But we may need to find a better way

voluncheer/opportunityboard/views/postajob.py
dbleek: Yes, yes it was. File is deleted.

voluncheer/opportunityboard/urls.py
dbleek: I changed the url from "post_an_opportunity" to "post." Full path is "opportunityboard/post."

voluncheer/opportunityboard/forms/postanopportunity.py
```diff
@@ -0,0 +1,54 @@
...
+        widget=forms.TimeInput(
+            attrs={
+                "type": "time",
+            }
+        ),
+    )
+
+    class Meta:
+        model = Opportunity
+        fields = (
```
joeleu, line 30: These are the required fields in the form, yes? If that's true, are we saying that the `date` and `duration` are not required?

Maybe something for the future... We should definitely make those fields required when an organization is publishing an opportunity to the public.

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -0,0 +1,18 @@
+from django.apps import apps
+from django.shortcuts import render
+
+from opportunityboard.models import Opportunity
+
+Organization = apps.get_model("profiles", "Organization")
```
joeleu, line 6: I'll believe you if you tell me this is "how it is done in Django" but it surprises me that this isn't just an import statement. Can you point me to the Django documentation for this pattern for my own edification?

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def opportunityboard(request):
```
joeleu, line 9: Public functions and classes should have docstrings, even if just a one-liner. Please add a docstring for these two functions.

voluncheer/profiles/test_forms.py
```diff
@@ -1,7 +1,21 @@
+from datetime import datetime
+
 from django.test import TestCase
 
 from profiles.forms.volunteers import VolunteerCreationForm
+from profiles.forms.volunteers import VolunteerChangeForm
+from profiles.forms.volunteers import VolunteerChangeForm
```
joeleu, line 7: I _think_ this is being imported twice... Same below for a number of other imports.

voluncheer/opportunityboard/forms/postanopportunity.py
dbleek: The fields attribute in a Model Form is just which fields from the model appear in the form. Date and duration required some customization, hence the declarations outside of `fields`. That said, I added them back into `fields ` as well, because that affects the order of how the fields appear and I prefer them higher up in the form.

voluncheer/opportunityboard/views/opportunityboard.py
dbleek: I don't think this line is necessary because `Organization `is not used in either of the functions. Eden could you please confirm?

voluncheer/opportunityboard/views/opportunityboard.py
dbleek: Will add docstrings. Not sure what the difference is between the two views, confirming with Eden separately.

voluncheer/profiles/test_forms.py
dbleek: Good catch, not sure why everything was doubled up. Fixed all the module imports and sorted them alphabetically.

voluncheer/opportunityboard/views/opportunityboard.py
dbleek: Confirmed not necessary.

PR #88, opened 2023-03-14
Password reset/Email Confirmation

Hgersten-hash:
This pr includes the tests for password_reset as well as the email verification implementation. It doesn't include tests yet for email verification.


voluncheer/profiles/test_views.py
```diff
@@ -1,10 +1,10 @@
 from django.test import TestCase
 from django.urls import reverse
 
-from profiles.models import Organization
-from profiles.models import User
-from profiles.models import UserType
-from profiles.models import Volunteer
+# from profiles.models import Organization
```
JinweiZhu, line 4: I think we should remove the imports if we decide to not use them?

voluncheer/profiles/urls.py
```diff
@@ -1,12 +1,13 @@
 from django.urls import path
-from django.urls import re_path
```
JinweiZhu, line 2: I thought you are supposed to use re_path, because otherwise you would get a warning for r"^profile/$", which is an outdated way? Maybe, my version was not correct when that warning showed up. So as long as you don't see any warning during migration and runserver, it should be good.

voluncheer/profiles/test_password_reset.py
```diff
@@ -0,0 +1,100 @@
...
+from django.urls import reverse
+from profiles.models import User
+from django.core import mail
+from django.utils.http import urlsafe_base64_encode
+from django.utils.encoding import force_bytes
+from django.contrib.auth.tokens import default_token_generator
+
+
+class PasswordResetTest(TestCase):
+    def setUp(self):
```
JinweiZhu, line 11: should it be set_up instead?

voluncheer/profiles/test_password_reset.py
JinweiZhu, line 1: In profiles directory of develop, there is already test_forms test_models and test_views.py files, would you want to move your tests there instead.

voluncheer/profiles/test_password_reset.py
```diff
@@ -0,0 +1,100 @@
...
+        self.token = default_token_generator.make_token(self.test_user)
+
+    def test_password_reset_form_view(self):
+        response = self.client.get(reverse("password_reset"))
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(
+            response, "registration/password_reset_form.html"
+        )  # noqa E501
+
+    def test_password_reset_form(self):
```
JinweiZhu, line 31: I don't quite get why you need to test for status code 200 and then 302? Is there a redirection happening only when email specified?

voluncheer/profiles/test_password_reset.py
```diff
@@ -0,0 +1,100 @@
...
+        response = self.client.get(
+            reverse("password_reset_confirm", args=[self.uidb64, self.token])
+        )  # noqa E501S
+        self.assertEqual(response.status_code, 302)
+        # extract the URL to which the user was redirected
+        redirect_url = response.url
+        # make a GET request to the redirect URL
+        response = self.client.post(
+            redirect_url,
+            {
```
JinweiZhu, line 81: Why password1 and password2? Did you inherent directly from Django?

voluncheer/profiles/views/volunteers.py
```diff
@@ -20,6 +21,17 @@ def get_context_data(self, **kwargs):
...
+        user.is_active = True
+        user.save()
+        volunteer_profile = Volunteer.objects.create(
+            user=user,
+            first_name=form.cleaned_data.get("first_name"),
+            last_name=form.cleaned_data.get("last_name"),
+            date_of_birth=form.cleaned_data.get("date_of_birth"),
+        )
+        volunteer_profile.save()
+        user.is_active = False
```
JinweiZhu, line 34: Why do you make the user inactive? It's like deleting the user right?

JinweiZhu: https://stackoverflow.com/questions/29742845/django-how-do-i-use-is-active-of-auth-user-table

voluncheer/profiles/views/organizations.py
```diff
@@ -20,6 +19,15 @@ def get_context_data(self, **kwargs):
...
-        return redirect("profile")
+        user = form.save(commit=False)
+        user.is_active = True
+        user.save()
+        organization_profile = Organization.objects.create(
+            user=user,
+            name=form.cleaned_data.get("name"),
+        )
+        organization_profile.save()
+        user.is_active = False
```
JinweiZhu, line 30: Same as below. I don't think you should touch is_active, unless it's a custom field.

JinweiZhu (changes requested):

voluncheer/profiles/views/volunteers.py
Hgersten-hash: This is for email verification purposes. We want the user to be unable to login unless they have verified their email successfuly

voluncheer/profiles/views/organizations.py
Hgersten-hash: Same as above.

voluncheer/profiles/test_password_reset.py
Hgersten-hash: This is to get the password and confirm password fields from the Reset Password page.

voluncheer/profiles/test_password_reset.py
Hgersten-hash: Yes, when email is specified this means that the user requested to send the reset password link to their email. The test_password_reset_form makes sure that the email is sent and that it contains the correct information inside. The next function we test to see that the link works for the user which is passed the uid and token arguments. That url redirects to the password reset page where they can enter their new password (and confirm their new password). The next function makes sure that once they enter that information, the page redirects to password_reset_complete, and that the password is updated in the database for the user. Finally, the last function tests that the password_reset_complete page is working and calling the correct template. Does this make sense?

voluncheer/profiles/test_password_reset.py
Hgersten-hash: I was thinking that it might be better to just have the tests separate and make one big test folder so we can manage them more easily.

voluncheer/profiles/test_views.py
Hgersten-hash: This isn't my addition, so I am not sure what they were intending. I agree with you, but I don't want to mess with someone else's code.

voluncheer/profiles/urls.py
Hgersten-hash: Same as above. I did not implement this code so I am not sure what the person was intending.

voluncheer/profiles/test_password_reset.py
Hgersten-hash: It can be but I am not sure if it matters. I can change it if you'd like!

voluncheer/profiles/test_password_reset.py
JinweiZhu: Okay, but are you willing to restructure the existing test files?

voluncheer/profiles/test_password_reset.py
dbleek: `setUp()` is a built-in testing function in Django. Basically, it is used to write data to the db temporarily before each test and clears itself after each test. For example, if you wanted a test user, you would create the user here, and you can rest assured that would you get a fresh user for each test (as to it being modified and passed along from test to test). 

This should remain as `setUp()`.

joeleu: `setUp` is the proper function, it is inherited from `unittest.TestCase` and pre-dates the PEP-8 style of using `lower_snake_case`.

voluncheer/profiles/test_password_reset.py
```diff
@@ -0,0 +1,100 @@
...
+        self.test_user.save()
+        self.uidb64 = urlsafe_base64_encode(force_bytes(self.test_user.pk))
+        self.token = default_token_generator.make_token(self.test_user)
+
+    def test_password_reset_form_view(self):
+        response = self.client.get(reverse("password_reset"))
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(
+            response, "registration/password_reset_form.html"
+        )  # noqa E501
```
joeleu, line 29: These lint and format exceptions should only be used as a last resort. I don't think this one is even necessary, is it?

voluncheer/profiles/test_password_reset.py
```diff
@@ -0,0 +1,100 @@
...
+            reverse("password_reset"), {"email": self.test_user.email}
+        )
+        response = self.client.get(
+            reverse("password_reset_confirm", args=[self.uidb64, self.token])
+        )  # noqa E501S
+        self.assertEqual(response.status_code, 302)
+        # extract the URL to which the user was redirected
+        redirect_url = response.url
+        # make a GET request to the redirect URL
+        response = self.client.get(redirect_url)
```
joeleu, line 62: Optional: you could just stick `response.url` in here instead of assigning it to a separate variable.

voluncheer/profiles/test_password_reset.py
```diff
@@ -0,0 +1,100 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(
+            response, "registration/password_reset_confirm.html"
+        )  # noqa E501
+
+    def test_password_reset_confirm_view(self):
+        response = self.client.post(
```
joeleu, line 69: This instance of `response` is never used since it is immediately overwritten on line 72, if that is intentional you can just not assign the return value to anything.

voluncheer/profiles/test_views.py
joeleu: We should definitely remove this before merging though. It's OK if this PR is still a work in progress but upon submission we should have commented out code.

For the rest of this review I'll assume all of the commented code falls into this.

voluncheer/profiles/urls.py
```diff
@@ -1,12 +1,13 @@
...
 
 from profiles.views.home import home
 from profiles.views.profile import ProfileView
 from profiles.views.profile import profile_update
+from profiles.views.activate_email import activate
```
joeleu, line 6: Please sort the local imports alphabetically. It makes parsing imports easier for us humans.

voluncheer/profiles/views/activate_email.py
```diff
@@ -0,0 +1,62 @@
...
+            "admin@admin.com",
+            [to_email],
+            fail_silently=False,  # noqa E501
+        )  # noqa E501
+        messages.success(
+            request,
+            f"Dear <b>{user}</b>, please go to your email <b>{to_email}</b> inbox and click on \
+        the received activation link to confirm and complete the registration. <b>Note:</b> Check your spam folder.",  # noqa E501
+        )
+    except BadHeaderError:
```
joeleu, line 39: This seems like a very specific exception type to catch, is this the "standard practice" in Django? Or did you run into this error? I'm wondering if there are other cases we want to catch.

voluncheer/profiles/views/activate_email.py
```diff
@@ -0,0 +1,62 @@
...
+        return HttpResponse("Invalid header found.")
+    return redirect("login")
+
+
+def activate(request, uidb64, token):
+    User = get_user_model()
```
joeleu, line 45: Fancy, does this return our custom user or the built-in Django one?

voluncheer/profiles/views/activate_email.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        uid = force_str(urlsafe_base64_decode(uidb64))
+        user = User.objects.get(pk=uid)
+    except (TypeError, ValueError, OverflowError, User.DoesNotExist):
+        user = None
+    if user is not None and token:
```
joeleu, line 51: nit: This is a good opportunity for a guard clause and a positive boolean check. By making boolean checks positive rather than negative we don't have to be concerned with potential double negatives. It's also just easier for us humans to read `if True` statements rather than `if False` statements.

voluncheer/profiles/views/activate_email.py
```diff
@@ -0,0 +1,62 @@
+from django.http import HttpResponse
```
joeleu, line 1: nit: please sort the imports alphabetically for us humans.

voluncheer/profiles/views/activate_email.py
```diff
@@ -0,0 +1,62 @@
...
+from django.template.loader import render_to_string
+from django.utils.http import urlsafe_base64_encode, urlsafe_base64_decode
+from django.contrib.auth.tokens import default_token_generator
+from django.utils.encoding import force_bytes, force_str
+from django.contrib import messages
+from django.contrib.auth import get_user_model
+from django.shortcuts import redirect
+
+
+def activateEmail(request, user, to_email):
```
joeleu, line 12: Public functions should have dostrings, even if they are just simple one-liners: https://peps.python.org/pep-0008/#documentation-strings

Here and elsewhere.

voluncheer/profiles/views/volunteers.py
joeleu: Do we have to set it to true, save the user, then set it back to false, and re-save the user? Is that a required step?

voluncheer/voluncheer/.env
joeleu, line 1: This file has been removed upstream, you can either remove it here or just sync your branch to the HEAD of develop.

voluncheer/voluncheer/settings.py
```diff
@@ -168,9 +168,9 @@
 EMAIL_BACKEND = "django_ses.SESBackend"
 AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY")
 AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
-AWS_SES_REGION_NAME = "us-west-2"  # (ex: us-east-2)
+AWS_SES_REGION_NAME = "us-east-1"  # (ex: us-east-2)
```
joeleu, line 171: Sadly, `us-east-1` doesn't have SES. I believe we have to use `us-east-2` for SES specifically. Also, please remove the comments that follow these lines, they don't really add anything.

joeleu (changes requested):

voluncheer/profiles/test_password_reset.py
Hgersten-hash: For some reason, until I used it every time I tried to restructure it on my own, after saving the code would go back to the original format I had. Oddly, using the exception restructured the code for me...

voluncheer/profiles/views/volunteers.py
Hgersten-hash: Yes, this is required because of the way we implemented our User types. Django now has an error: "Assigning unsaved objects to a ForeignKey, GenericForeignKey, and OneToOneField now raises a ValueError.". When we try to save the user with commit=false, it raised an error when we tried to save the user because of the fact that Volunteer/Organizations use that foreign key. Also, if I tried to only create the Volunteer/Organization profile after the activation link, the data from the form would be lost. To solve this, I first set the user to active so that the organization/volunteer profile could be created without issue. Then I save the organization/volunteer profile and then I save the user; finally I set the is_active to false so that the user must activated their account to be able to login.

PR #91, opened 2023-03-14
Add post-submit CI for the main branch

joeleu:
I noticed the assignment asks to specifically run a daily test of the `main` branch so that's what this does.

The next step would be to move this to a re-usable workflow and then reference that in the various configurations but we can do that later.


dbleek (approved):
All looks good to me! I compared to the pre-submit workflow and made sure any changes made sense, which they do.

PR #93, opened 2023-03-15
Cleaned up urls.py in project root folder

dbleek:
There were a few URLs related to the profiles app that I moved into that app's namespace. I also sorted the paths the settings.py file in the project root folder and simplified the URLs (mostly removing "account/"). These changes will make the codebase slightly more organized and the URLs cleaner.


voluncheer/voluncheer/urls.py
```diff
@@ -18,28 +18,14 @@
...
-from profiles.views.home import SignUpView
-from profiles.views.organizations import OrganizationSignUpView
-from profiles.views.volunteers import VolunteerSignUpView
 
 urlpatterns = [
-    path("__debug__/", include("debug_toolbar.urls")),
-    path("admin/", admin.site.urls),
+    path("", include("django.contrib.auth.urls")),
     path("", include("profiles.urls")),
-    path("accounts/", include("django.contrib.auth.urls")),
```
joeleu, line 29: Using the `accounts` prefix is what the django documentation had as an example. They don't really go into detail as to why but I assume to avoid overload/conflict. As long as we OK not using those url patterns, this is fine.

voluncheer/profiles/urls.py
```diff
@@ -2,11 +2,25 @@
...
 from profiles.views.home import home
+from profiles.views.home import SignUpView
+from profiles.views.organizations import OrganizationSignUpView
 from profiles.views.profile import ProfileView
 from profiles.views.profile import profile_update
+from profiles.views.volunteers import VolunteerSignUpView
 
 urlpatterns = [
     path("", home, name="home"),
     path(r"^profile/$", ProfileView.as_view(), name="profile"),
```
joeleu, line 13: If you're cleaning this up please remove the regex anchors as they are not be necessary. Here and below.

voluncheer/voluncheer/urls.py
```diff
@@ -18,28 +18,14 @@
...
-        "accounts/signup/volunteer/",
-        VolunteerSignUpView.as_view(),
-        name="volunteer_signup",
-    ),
+    path("admin/", admin.site.urls),
+    path("__debug__/", include("debug_toolbar.urls")),
+    path("map/", include("map.urls")),
+    path("opportunityboard/", include("opportunityboard.urls")),
     # Unimplemented urls.
     # path("chat/", include("chatroom.urls")),
```
joeleu, line 30: Please remove these commented out lines.

voluncheer/voluncheer/urls.py
```diff
@@ -18,28 +18,14 @@
...
-        "accounts/signup/organization/",
-        OrganizationSignUpView.as_view(),
-        name="organization_signup",
-    ),
-    path(
-        "accounts/signup/volunteer/",
-        VolunteerSignUpView.as_view(),
-        name="volunteer_signup",
-    ),
+    path("admin/", admin.site.urls),
```
joeleu, line 25: Odd that this one isn't in an `include()` call. Would you happen to know why that is?

joeleu (changes requested):

voluncheer/profiles/urls.py
JinweiZhu: Same

voluncheer/voluncheer/urls.py
```diff
@@ -18,28 +18,14 @@
 from django.urls import path
 from django.contrib.auth import views as auth_views
 
-from profiles.views.home import SignUpView
-from profiles.views.organizations import OrganizationSignUpView
-from profiles.views.volunteers import VolunteerSignUpView
 
 urlpatterns = [
-    path("__debug__/", include("debug_toolbar.urls")),
```
JinweiZhu, line 26: I think we won't have debug toolbar if we remove this?

voluncheer/profiles/urls.py
dbleek: Done.

voluncheer/voluncheer/urls.py
dbleek: Done.

voluncheer/voluncheer/urls.py
dbleek: Just to be safe, I put the `accounts` back to the URLConf that leads to authentication URLConf. But it shouldn't be necessary for any of the signup views--this URLConf is simply for login/logout/password reset/password change.

voluncheer/voluncheer/urls.py
dbleek: For whatever reason, it's not necessary. See: https://docs.djangoproject.com/en/4.1/ref/contrib/admin/#hooking-adminsite-to-urlconf

joeleu: /shrug I guess we have to just "take Django's word for it".

joeleu (approved):
LGTM

voluncheer/profiles/test_views.py
```diff
@@ -12,7 +12,7 @@ def test_signup_form_accessible_by_name(self):
 
     def test_signup_form_exists_at_desired_location(self):
         """Tests accessing  signup form by URL"""
-        response = self.client.get("/accounts/signup/")
```
JinweiZhu, line 15: I think unless you change LOGIN_REDIRECT_URL in settings.py, we should have have URL starting with accounts:
https://stackoverflow.com/questions/13186693/why-second-user-login-redirects-me-to-accounts-profile-url
Is the URL accessible after the change?

joeleu:
I think this PR is obsolete now. Doug, is that the case? If so, can you close this PR.

dbleek:
Will be addressed by profile_url_cleanup

PR #94, opened 2023-03-15
Password reset/Email Confirmation

Hgersten-hash:
[This pr includes the tests for password_reset as well as the email verification implementation. It doesn't include tests yet for email verification


voluncheer/static/fonts/ZenKakuGothicNew-Black.ttf
EdenWuyifan, line 1: Those fonts are used for our pages. Don't delete them ⛔

voluncheer/voluncheer/wsgi.py
EdenWuyifan, line 1: Is there a reason why you deleted wsgi.py? I think it will cause problems while doing runserver

PR #95, opened 2023-03-16
Add flake8 and black configuration

joeleu:
This PR adds configuration files for both flake8 and black. Those files are then set in the `.vscode/settings.json` to ensure the editor and its plugins use the same configuration. This makes sure both the CI/CD environment and our local environments have the same linting and formatting expectations.

Finally, the Python related changes are due to these changes in the configuration and are formatting only.


voluncheer/.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+max-line-length = 100
+exclude = */migrations/*
```
JinweiZhu, line 3: *_initial.py should also be included? Those migration files are generated by the system.

JinweiZhu (changes requested):
Minor change is needed but otherwise looks good.

voluncheer/.flake8
joeleu: The only `*_initial.py` files I saw were within a `migrations/` directory so they are covered by this exclude. Do these files appear elsewhere?

JinweiZhu: Good point.

JinweiZhu (approved):

PR #99, opened 2023-03-16
Email Configurations for AWS

Hgersten-hash:
Updated configurations for AWS and made everything environment variables so that we don't need to deploy every time.


voluncheer/profiles/views/activate_email.py
```diff
@@ -18,9 +18,7 @@ def activateEmail(request, user, to_email):
         {
             "email": user.email,
             "user": user,
-            "domain": "voluncheer-main.us-east-1.elasticbeanstalk.com"  # noqa E501
-            if os.getenv("IS_PRODUCTION")
-            else "127.0.0.1:8000",  # noqa E501
+            "domain": os.getenv("DOMAIN"),  # noqa E501
```
joeleu, line 21: These `noqa E501` comments should be removed.

joeleu, line 21: Add a prefix to the environment variable related to the use of the variable, for instance `SES_SENDER_DOMAIN` or something like that. That way when we are _only_ looking at the environment variable we can tell what it is used for.

We should also pull these environment variables when the file is initialized (or ideally in a shared file during initialization) to avoid having to read them every time these functions are used. By reading them during initialization they are only read once and then re-used as long as the server remains alive. All in all it is much more efficient.

Here and elsewhere.

voluncheer/requirements.txt
```diff
@@ -27,4 +27,4 @@ six==1.16.0
 sqlparse==0.4.3
 urllib3==1.26.14
 verify-email==2.4.3
-Pillow==9.3.0
+Pillow==9.3.0
```
joeleu, line 30: I don't think `Pillow` is used anymore. This can be removed. It is from an old commit.

joeleu (changes requested):

voluncheer/requirements.txt
joeleu: I was wrong about this, leave it as is.

voluncheer/profiles/views/activate_email.py
```diff
@@ -31,7 +29,7 @@ def activateEmail(request, user, to_email):
         send_mail(
             subject,
             message,
-            "admin@admin.com",
+            os.getenv("EMAIL"),
```
joeleu, line 32: Is this the "from" email? Can you update this one to be more descriptive? Here and in `profile.py`.

voluncheer/voluncheer/settings.py
```diff
@@ -165,14 +165,15 @@
...
-    EMAIL_BACKEND = "django_ses.SESBackend"
-    AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY")
-    AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
-    AWS_SES_REGION_NAME = "us-east-2"
-    AWS_SES_REGION_ENDPOINT = "email.us-east-2.amazonaws.com"
-else:
-    EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
+# if os.getenv("IS_PRODUCTION"):
+EMAIL_BACKEND = "django_ses.SESBackend"
+AWS_ACCESS_KEY_ID = os.getenv("SES_AWS_ACCESS_KEY")
```
joeleu, line 170:
```suggestion
AWS_SES_ACCESS_KEY_ID = os.getenv("AWS_SES_ACCESS_KEY_ID")
```

https://github.com/django-ses/django-ses#getting-going

voluncheer/voluncheer/settings.py
```diff
@@ -165,14 +165,15 @@
...
+AWS_SECRET_ACCESS_KEY = os.getenv("SES_AWS_SECRET_ACCESS_KEY")
```
joeleu, line 171:
```suggestion
AWS_SES_SECRET_ACCESS_KEY = os.getenv("AWS_SES_SECRET_ACCESS_KEY")
```

https://github.com/django-ses/django-ses#getting-going

voluncheer/voluncheer/settings.py
```diff
@@ -165,14 +165,15 @@
...
+AWS_SES_REGION_NAME = os.getenv("SES_REGION_NAME")
```
joeleu, line 172:
```suggestion
AWS_SES_REGION_NAME = os.getenv("AWS_SES_REGION_NAME")
```

voluncheer/voluncheer/settings.py
```diff
@@ -165,14 +165,15 @@
...
+AWS_SES_REGION_ENDPOINT = os.getenv("SES_REGION_ENDPOINT")
```
joeleu, line 173:
```suggestion
AWS_SES_REGION_ENDPOINT = os.getenv("AWS_SES_REGION_ENDPOINT")
```

joeleu (changes requested):

voluncheer/profiles/views/activate_email.py
joeleu: For posterity, these will get cleaned up a bit later.

voluncheer/profiles/views/activate_email.py
joeleu: This was updated in later commits.

voluncheer/profiles/views/activate_email.py
joeleu: One thing still not done is moving these environment reads from within the function to module initialization.

joeleu: If you don't want to do it now at least add a TODO comment so we can keep track of what we still want to improve.

Hgersten-hash: Never mind I think I have an idea!

voluncheer/voluncheer/settings.py
```diff
@@ -140,12 +140,12 @@
...
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/4.1/howto/static-files/
-STATIC_ROOT = os.path.join(BASE_DIR, "static")
+# STATIC_ROOT = os.path.join(BASE_DIR, "static")
 STATIC_URL = "/static/"
 # The following lines should be added while test locally.
-# STATICFILES_DIRS = [
-#     os.path.join(BASE_DIR, "static"),
-# ]
+STATICFILES_DIRS = [
```
joeleu, line 146: Does this need to be commented out for deployment on AWS? Opposite above for the `STATIC_ROOT`?

voluncheer/profiles/views/activate_email.py
joeleu: Yes, perfect.

voluncheer/voluncheer/settings.py
Hgersten-hash: Yes, thank you!

joeleu (approved):
LGTM. Ship it!

PR #100, opened 2023-03-17
Add auto-sorting of imports

joeleu:
This PR adds auto-sorting of imports through `isort`, configures it for integration with vscode, and checks import sorting during CI/CD runs. Finally, this sorts the existing imports within the respository.


JinweiZhu (approved):
LGTM.

PR #101, opened 2023-03-19
Subcategories

dbleek:
This implements Categories, Subcategories, and Subsubcategories which organization can select when posting an opportunity, as well as some minor tweaks to the opportunity model for clarity.

All categories are stored in the db to allow for creation/deletion via the Django admin.

If possible, I would like to eventually refactor this so that all of the above fall under Categories, or at least Categories and Subcategories for less repetition in the code.

Testing for forms and views to come as well.


voluncheer/opportunityboard/forms/postanopportunity.py
```diff
@@ -26,30 +27,77 @@ class PostAnOpportunityForm(forms.ModelForm):
...
+        self.fields["subsubcategory"].queryset = Subsubcategory.objects.none()
+
+        if "category" in self.data:
+            try:
+                parent_id = int(self.data.get("category"))
+                self.fields["subcategory"].queryset = Subcategory.objects.filter(
+                    parent_id=parent_id
+                ).order_by("name")
+            except (ValueError, TypeError):
+                pass
```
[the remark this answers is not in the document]
JinweiZhu: Same as below.

voluncheer/opportunityboard/forms/postanopportunity.py
```diff
@@ -26,30 +27,77 @@ class PostAnOpportunityForm(forms.ModelForm):
...
+        labels = {"subcategory": "Subcategory 1", "subsubcategory": "Subcategory 2"}
+
+        widgets = {"organization": forms.HiddenInput()}
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["subcategory"].queryset = Subcategory.objects.none()
+        self.fields["subsubcategory"].queryset = Subsubcategory.objects.none()
+
+        if "category" in self.data:
```
JinweiZhu, line 56: Are you trying to conduct test here? What exactly are you trying to accomplish?

dbleek: We need to override `__init__` for the form to render properly. For example, when a form for a new opportunity loads, the subcategories drop-down should have no options because an opportunity hasn't been selected yet. We then populate the drop down with a list of subcategories based on the category selection.

That particular test is if we would reload to the same form page after submitting it, which we are not doing currently. For us, after submitting the form, we redirect to the org. profile page. So theoretically we could probably delete everything up to the `elif`, but it's also not hurting us and covers us in case we decide to change down the road. Good catch, I'm open to either option.

See this tutorial for reference: https://simpleisbetterthancomplex.com/tutorial/2018/01/29/how-to-implement-dependent-or-chained-dropdown-list-with-django.html

voluncheer/opportunityboard/test_models.py
```diff
@@ -11,9 +14,10 @@
 class OpportunityTest(TestCase):
     """Test cases for Opportunity model"""
 
-    def setUp(self):
+    @classmethod
+    def setUpTestData(cls):
         """See base class."""
-        self.jedi = Organization.objects.create(
+        cls.jedi = Organization.objects.create(
```
JinweiZhu, line 20: This should be cls.org instead of clas.jedi

voluncheer/opportunityboard/test_models.py
```diff
@@ -22,13 +26,18 @@ def setUp(self):
             name="Jedi Council",
         )
 
-    def test_opportunity_details(self):
-        """Test basic opportunity details"""
+        cls.environment = Category.objects.create(name="Environment")
```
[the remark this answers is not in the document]
JinweiZhu: Similar below.

voluncheer/opportunityboard/test_models.py
```diff
@@ -52,15 +63,72 @@ def test_opportunity_details(self):
...
+
+    def test_subsubcategory_details(self):
+        """Test basic subsubcategory details"""
+        subsubcategory = self.reforestation
+        self.assertEqual(subsubcategory.name, "Reforestation")
+        self.assertEqual(subsubcategory.parent.name, "Conservation")
+
+    def test_opportunity_object_name_is_title(self):
+        """Tests Opportunity __str__ method returns title"""
+        opportunity = Opportunity.objects.get(id=1)
```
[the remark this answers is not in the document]
JinweiZhu: Same below.

voluncheer/opportunityboard/tests_views.py
```diff
@@ -1,10 +1,69 @@
...
+from opportunityboard.models import Subsubcategory
+from profiles.models import Organization
+from profiles.models import User
+from profiles.models import UserType
+
 
 class OpportunityboardTestCase(TestCase):
     """Test cases for Opportunityboard view"""
 
+    @classmethod
```
JinweiZhu, line 18: Why is this looking just alike test_model above? Isn't it redundant?

voluncheer/opportunityboard/views/postanopportunity.py
```diff
@@ -13,15 +17,20 @@ def post_an_opportunity(request):
...
+        form = PostAnOpportunityForm(request.POST, request.FILES)
+        # Save
         if form.is_valid():
             form.save()
         else:
             print(form.errors.as_data())
         return redirect("home")
 
     else:
-        opportunity_form = PostAnOpportunityForm(instance=request.user)
```
JinweiZhu, line 24: Why is the instance=Model no longer needed here?

voluncheer/opportunityboard/views/postanopportunity.py
```diff
@@ -13,15 +17,20 @@ def post_an_opportunity(request):
     if user.is_organization:
         organization_profile = Organization.objects.get(pk=user)
     if request.method == "POST":
-        form = PostAnOpportunityForm(request.POST, request.FILES, instance=request.user)
+        # Set Organization based on current user
+        post = request.POST.copy()
+        post.update({"organization": organization_profile})
```
JinweiZhu, line 22: Why do you have to copy and then update? Why can't you directly update?

JinweiZhu (changes requested):
Good work. Almost there.

voluncheer/opportunityboard/test_models.py
dbleek: Same as above

voluncheer/opportunityboard/tests_views.py
dbleek: Because the test data from test_models doesn't carry over to test_views, we need to recreate it. In general, we should expect some repetition in unit tests.

voluncheer/opportunityboard/views/postanopportunity.py
dbleek: Because the organization field is hidden in the form. Users shouldn't have to specify which organization they are, it should happen automatically in the code. This relates to above change (removing `instance=request.user`)

Hgersten-hash:
Just tested it out locally and it looks great! One question: Should we make the first general category required?

voluncheer/opportunityboard/test_models.py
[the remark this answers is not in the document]
JinweiZhu: Right, but according to OOP, it should be parent.child right? Right now, it's quite hard to read.

voluncheer/opportunityboard/views/postanopportunity.py
JinweiZhu: So you cannot do: request.POST.update({"organization": organization_profile}) ?

voluncheer/opportunityboard/forms/postanopportunity.py
JinweiZhu: Let's keep it like this for right now since it's working.

voluncheer/opportunityboard/test_models.py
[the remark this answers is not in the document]
JinweiZhu: Right, but you don't know how in the future if anyone is going to insert anything before? A best approach would be to get by email or sth.

voluncheer/opportunityboard/views/postanopportunity.py
dbleek: My understanding is you have to copy `request.POST` to make it mutable, but I'll double check

dbleek:
> Just tested it out locally and it looks great! One question: Should we make the first general category required?

Yes, better validation is something I was thinking about. That should be an easy change

voluncheer/opportunityboard/test_models.py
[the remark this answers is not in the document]
dbleek: Changed.

voluncheer/opportunityboard/tests_views.py
dbleek: Since the test db is made from scratch, I needed at least one opportunity in order to test the `update_an_opportunity` url, which is why I simply copied the code over. I'll include a comment to clarify

voluncheer/opportunityboard/views/postanopportunity.py
dbleek: I confirmed you have to make a copy before updating it. I agree it's weird, but necessary.

voluncheer/opportunityboard/test_models.py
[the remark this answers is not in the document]
dbleek: I changed all tests where we were using pk to `get `the opportunity.

JinweiZhu (approved):
It's ready. Good work.

PR #102, opened 2023-03-19
Default_From_Email 

Hgersten-hash:
Changed from email variable to override django's built in from email


EdenWuyifan (commented):
Good good

PR #103, opened 2023-03-19
Update opportunities, upload opportunity feature picture 

EdenWuyifan:


voluncheer/opportunityboard/models.py
dbleek, line 1: As we discussed today, I don't think staffing should be set to null=True, blank=True. This will be addressed when I merge the subcategories branch.

voluncheer/opportunityboard/tests_views.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def test_select_page_loads(self):
+        """Tests select page loads"""
+        response = self.client.get(reverse("select"))
+        self.assertEqual(response.status_code, 200)
+
+    def test_post_an_opportunity_page_loads(self):
+        """Tests post_an_opportunity page loads"""
+        response = self.client.get(reverse("post_an_opportunity"))
+        self.assertIn(response.status_code, [200, 302])
```
dbleek, line 21: Why are we allowing for success or redirect? Same for below.

voluncheer/opportunityboard/tests_views.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def test_update_an_opportunity_page_loads(self):
+        """Tests update_an_opportunity page loads"""
+        response = self.client.get(reverse("update_an_opportunity", args=[1]))
+        self.assertIn(response.status_code, [200, 302])
```
dbleek, line 26: You might have to create an opportunity in the test setUp for this to work correctly. You could copy it from the models test.

voluncheer/opportunityboard/views/postanopportunity.py
```diff
@@ -2,22 +2,50 @@
...
+    if user.is_anonymous:
+        return redirect("home")
+    if user.is_organization:
+        organization_profile = Organization.objects.get(pk=user)
     if request.method == "POST":
-        form = PostAnOpportunityForm(request.POST, instance=request.user)
+        form = PostAnOpportunityForm(request.POST, request.FILES, instance=request.user)
         if form.is_valid():
             form.save()
             return redirect("home")
```
dbleek, line 19: We should have an `else `branch here in case the form is not valid. I'm not sure if that is just a matter of raising an error or perhaps displaying an error message on the page.

voluncheer/opportunityboard/views/postanopportunity.py
```diff
@@ -2,22 +2,50 @@
...
+    user = request.user
+    if user.is_anonymous:
+        return redirect("home")
+    if user.is_organization:
+        organization_profile = Organization.objects.get(pk=user)
+    if request.method == "POST":
+        form = PostAnOpportunityForm(request.POST, request.FILES, instance=request.user)
+        if form.is_valid():
+            form.update(opportunity_id)
+            return redirect("home")
```
dbleek, line 40: Same as above.

voluncheer/voluncheer/settings.py
dbleek, line 1: Not sure why these lines were moved down or why we are using `"media"` instead of `MEDIA_URL`. Are these changes necessary?

voluncheer/opportunityboard/test_forms.py
```diff
@@ -0,0 +1,69 @@
...
+            "description": "Face your destiny.",
+            "date": timezone.now(),
+            "duration": timedelta(minutes=20),
+            "address_1": "New York, NY",
+            "address_2": "Mean st",
+            "is_published": True,
+        }
+        form = PostAnOpportunityForm(data=data, instance=self.user)
+        self.assertEqual(self.user.opportunity_set.all().count(), 0)
+        self.assertFalse(form.save())
```
dbleek, line 43: What does it mean to assert that `form.save()` is `False`? Same with the line in the below that asserts if `form.update(1)` is `False`.

voluncheer/opportunityboard/test_forms.py
```diff
@@ -0,0 +1,69 @@
...
+                password="peace_and_justice_for_the_galaxy",
+                type=UserType.ORGANIZATION,
+            ),
+            name="Jedi Council",
+        )
+        self.user.save()
+
+    def test_validation(self):
+        """Test Post An Opportunity validation"""
+        data = {
```
dbleek, line 31: This can go into a built-in Django test function called `setUp(self)`, which create an opportunity on the test db for each test to use.

voluncheer/opportunityboard/test_forms.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertFalse(form.save())
+        self.assertEqual(self.user.opportunity_set.all().count(), 1)
+        posted_opportunity = self.user.opportunity_set.all().first()
+        self.assertEqual(posted_opportunity.title, "Jedi Trial")
+        self.assertEqual(posted_opportunity.category, "sports")
+        self.assertEqual(posted_opportunity.description, "Face your destiny.")
+
+    def test_update(self):
+        """Test Update An Opportunity validation"""
+        self.test_validation()
```
dbleek, line 52: Using a test in another test seems odd, are you sure we want this?

dbleek (changes requested):
Looks good! I noticed a couple of things I'm not sure about, please take a look and let me know if you have any questions.

voluncheer/opportunityboard/test_forms.py
EdenWuyifan: Yeah. I'll try add setUp

voluncheer/voluncheer/settings.py
EdenWuyifan: Yes, those are necessary. There are accessibility issues when uploading image locally, so I debugged it

voluncheer/opportunityboard/test_forms.py
EdenWuyifan: image

Void function can only be tested using assertFalse

voluncheer/opportunityboard/tests_views.py
EdenWuyifan: 302 error does not mean webpage couldn’t be loaded. It's like too many redirections or so... Hence I omit it

voluncheer/opportunityboard/models.py
EdenWuyifan: Sure thing. Feel free to over write that with your branch~

dbleek (approved):

PR #116, opened 2023-03-21
Add an environment class with expressive typing.

joeleu:
This PR adds a class and default for presenting our environment in a more expressive way. It provides easy to use booleans where we can branch specific executions (e.g. the static and media settings). Ideally this will make our various environments easier to use.

### DIFF
--- a/voluncheer/voluncheer/settings.py
+++ b/voluncheer/voluncheer/settings.py
@@ -13,6 +13,8 @@ https://docs.djangoproject.com/en/4.1/ref/settings/
 import os
 from pathlib import Path
 
+from voluncheer.environment import environment
+
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = Path(__file__).resolve().parent.parent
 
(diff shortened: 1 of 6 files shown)


joeleu:
*sigh* I'll address the tests tomorrow.

Hgersten-hash (commented):
This is a smart way to enable three different environment types. As was indicated by the test failure, we need to make it work with sending emails. I think this is because, in your code, the AWS_SES_DOMAIN isn't defined when the environment is local. Maybe, we can have it defined regardless of the environment type, and then check if it's empty in the domain part of the emails?

voluncheer/voluncheer/settings.py
```diff
@@ -153,34 +155,34 @@
...
+if not environment.is_local:
+    EMAIL_BACKEND = "django_ses.SESBackend"
+    AWS_ACCESS_KEY_ID = os.getenv("AWS_SES_ACCESS_KEY")
+    AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SES_SECRET_ACCESS_KEY")
+    AWS_SES_REGION_NAME = os.getenv("AWS_SES_REGION_NAME")
+    AWS_SES_REGION_ENDPOINT = os.getenv("AWS_SES_REGION_ENDPOINT")
+    AWS_SES_DOMAIN = os.getenv("AWS_SES_DOMAIN")
+    DEFAULT_FROM_EMAIL = os.getenv("AWS_SES_FROM_EMAIL")
+else:
+    EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
```
JinweiZhu, line 168: Since the test is failing for import, we can just use placeholder values like "N/A".

joeleu: An empty string is as useful as nonsense so I left it empty. We can't really test email locally right now anyway.

PR #121, opened 2023-03-22
added delete button and save progress when update

JinweiZhu:
Now delete button is on the opportunity board. Also, when you click update, the existing data is shown automatically. I didn't write tests because these are UI change. But if we should let me know. Also, failing CI due to black doesn't seem too right for me when we can correct it immediately.


JinweiZhu:
You have 2 buttons, 1 for update 1 for delete.

Hgersten-hash:
> You have 2 buttons, 1 for update 1 for delete.

I don't think we are putting these on the opportunity board. They belong on the editing form of posting an opportunity. Does that make sense?

JinweiZhu:
I don't think it makes a difference. We can keep it in for right now?

Hgersten-hash:
> I don't think it makes a difference. We can keep it in for right now?

I would rather remove them since Eden is working on the filtering part of the opportunity board and I wouldn't want this to affect that.

JinweiZhu:
I don't think the delete button will have any impact on the filtering. Besides, I don't find having delete right next to the pencil for edit a good UI layout. It's prone to misclick the icon.

dbleek (commented):
Looks great! Mostly had to sift through my own changes because of the rebase :) but deletion and loading previous opportunities looks good.

voluncheer/nuke-and-rebuild-db.sh
```diff
@@ -0,0 +1,18 @@
...
+find . -path "*/migrations/*pycache*"  -delete
+rm db.sqlite3
+#pip uninstall -y Django
+#pip install Django==3.2.8
+git checkout origin/develop -- db.sqlite3
+mv db.sqlite3 dbTemp.sqlite3
+./manage.py makemigrations
+./manage.py migrate
+#Dump data
+sqlite3 dbTemp.sqlite3 ".dump opportunityboard_category" |sqlite3 db.sqlite3
```
dbleek, line 14: Could you explain how this works? Just for my own edification.

EdenWuyifan:
Great work done!!! Thanks for the optimizations, Wayne!!

PR #123, opened 2023-03-23
Retrieve the Google Maps API Key from an environment variable

joeleu:
This environment variable is already set in both our AWS environments.


Hgersten-hash (approved):
Looks good to me! For localhost, we will all need to use our own API key which means we will need to have an env file I assume

PR #124, opened 2023-03-23
Update coveralls to use the v2 action

joeleu:
Using the `v2` GitHub action version is supposed to fix the error we are getting on post submit runs on main:

```
Error: Command failed: git cat-file -p 54f99fbb185008af3eae35e04dfb5e420b19a2a1 fatal: Not a valid object name 54f99fbb185008af3eae35e04dfb5e420b19a2a1
```


Hgersten-hash (approved):
Looks good. Thanks for taking care of this!

PR #125, opened 2023-03-24
Add filter search frontend, fitted the subcategory models

EdenWuyifan:
image

### DIFF
--- a/voluncheer/opportunityboard/tests_views.py
+++ b/voluncheer/opportunityboard/tests_views.py
@@ -1,12 +1,14 @@
 import datetime
 
 from django.test import TestCase
+from django.test.client import RequestFactory
 from django.urls import reverse
 
 from opportunityboard.models import Category
 from opportunityboard.models import Opportunity
 from opportunityboard.models import Subcategory
 from opportunityboard.models import Subsubcategory
+from opportunityboard.views.search import filter_search
 from profiles.models import Organization
 from profiles.models import User
 from profiles.models import UserType
@@ -78,3 +80,19 @@ class OpportunityboardTestCase(TestCase):
         """Tests update_an_opportunity page loads"""
         response = self.client.get(reverse("update_an_opportunity", args=[1]))
         self.assertIn(response.status_code, [200, 302])
+
+    def test_search(self):
+        """Tests update_an_opportunity page loads"""
+        rf = RequestFactory()
+        post_request = rf.post(
+            "/opportunityboard/search",
+            {
+                "category": "Animals",
+                "duration": "<=2 Hours",
+                "choices-single-defaul": "FUSION",
+                "distance": "2.9",
+                "startdates": "03/23/2023 - 03/23/2023",
+            },
+        )
+        response = filter_search(post_request)
+        self.assertEqual(response.status_code, 200)
--- a/voluncheer/opportunityboard/views/search.py
+++ b/voluncheer/opportunityboard/views/search.py
@@ -1,0 +1,60 @@
+import logging
+
+from django.shortcuts import render
+
+from opportunityboard.models import Category
+from opportunityboard.models import Opportunity
+from opportunityboard.models import Subcategory
+from opportunityboard.models import Subsubcategory
+from opportunityboard.views.opportunityboard import category_dict_gen
+
+
+def filter_search(request):
+    if request.method == "POST":
+        filter = parse_search_filter(request.POST)
+        logger = logging.getLogger(__name__)
+        logger.error(filter)
+
+    opportunity_lists = Opportunity.objects.order_by("-pubdate"[:20])
+    cate_output_dict = category_dict_gen()
+    durations = {
+        "One-day": ["<=2 Hours", "2-4 Hours", "Full Day"],
+    }
+    context = {
+        "opportunity_lists": opportunity_lists,
+        "categories": cate_output_dict,
+        "durations": durations,
+    }
+    return render(request, "voluncheer/opportunityboard.html", context)
+
+
+def parse_search_filter(post):
+    """Check if the input filters are valid:
+    category(default empty string)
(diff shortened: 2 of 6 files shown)


voluncheer/opportunityboard/views/search.py
```diff
@@ -0,0 +1,70 @@
...
+    subcategory = Subcategory.objects.filter(name=value).first()
+    if not subcategory:
+        return None, subcategory
+    subsubcategory = Subsubcategory.objects.filter(name=value).first()
+    if not subsubcategory:
+        return None, subsubcategory
+    return subsubcategory, None
+
+
+def duration_is_valid(value):
```
dbleek, line 53: The search errors out if you don't include a duration at all. Users may want to search for category without specifying a duration

dbleek (changes requested):

voluncheer/opportunityboard/views/search.py
```diff
@@ -0,0 +1,70 @@
...
+from opportunityboard.models import Subcategory
+from opportunityboard.models import Subsubcategory
+from opportunityboard.views.opportunityboard import category_dict_gen
+
+
+def filter_search(request):
+    if request.method == "POST":
+        err, category = category_is_valid(request.POST["category"])
+        if err is not None:
+            print(err)
```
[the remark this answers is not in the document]
joeleu: +1 please don't print errors to the console, those go nowhere. We should either surface them to the caller or discard them.

voluncheer/opportunityboard/views/search.py
joeleu: Also please don't leave code in that is either not in use or is commented out. Stash it locally or move it to another commit if you want to keep it around.

Here and below.

joeleu (changes requested):

voluncheer/opportunityboard/tests_views.py
```diff
@@ -78,3 +80,21 @@ def test_update_an_opportunity_page_loads(self):
         """Tests update_an_opportunity page loads"""
         response = self.client.get(reverse("update_an_opportunity", args=[1]))
         self.assertIn(response.status_code, [200, 302])
+
+    def test_search(self):
+        """Tests update_an_opportunity page loads"""
+        ordinary_dict = {
+            "csrfmiddlewaretoken": [
+                "rHoEJiKhW4ka5ycloy2pKcuEO5dPQw0nsW7thvfIxGEr4BAwOTcgyss6Obtvfhby",
```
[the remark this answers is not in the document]
JinweiZhu: Good point.

voluncheer/opportunityboard/views/search.py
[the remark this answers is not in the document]
JinweiZhu: Just throw a django exception or warning.

JinweiZhu (changes requested):
I think Doug and Joe have a few good points, but otherwise it looks good.

voluncheer/opportunityboard/views/search.py
```diff
@@ -0,0 +1,61 @@
...
+    duration(default empty string)
+    distance(default 0)
+    """
+    output = {}
+    try:
+        _, category = category_is_valid(post["category"])
+        output["category"] = category
+    except KeyError:
+        output["category"] = ""
+        pass
```
joeleu, line 43: `pass` doesn't add anything here.

voluncheer/opportunityboard/views/search.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def category_is_valid(value):
+    """Check if the category is valid
+    Input: value[category_name] -> string
+    Return: error(None), target_object
+    """
+    category = Category.objects.filter(name=value).first()
+    if not category:
+        return None, category
```
joeleu, line 54: This means that we are always returning a falsey `target_object`, right? Is that desirable? Maybe I don't understand what this function is trying to do.

voluncheer/opportunityboard/views/search.py
```diff
@@ -0,0 +1,61 @@
...
+
+def parse_search_filter(post):
+    """Check if the input filters are valid:
+    category(default empty string)
+    duration(default empty string)
+    distance(default 0)
+    """
+    output = {}
+    try:
+        _, category = category_is_valid(post["category"])
```
joeleu, line 39: if you use `post.get("category")` you won't need this error handling.

voluncheer/opportunityboard/views/search.py
```diff
@@ -0,0 +1,61 @@
...
+from opportunityboard.models import Category
+from opportunityboard.models import Opportunity
+from opportunityboard.models import Subcategory
+from opportunityboard.models import Subsubcategory
+from opportunityboard.views.opportunityboard import category_dict_gen
+
+
+def filter_search(request):
+    if request.method == "POST":
+        filter = parse_search_filter(request.POST)
```
joeleu, line 14: It's not clear to me what the `if request.method == "POST"` branch is trying to accomplish. Can you provide a docstring or some kind of indication as to why we are logging an error when the request is a post?

joeleu (changes requested):

voluncheer/opportunityboard/views/search.py
EdenWuyifan: this will be useful as protection after introducing filter search functions. For now it's just a log dump

voluncheer/opportunityboard/views/search.py
EdenWuyifan: Just an error handling when we can't find a match in Category, Subcategory, and Subsubcategory.

voluncheer/db.sqlite3
JinweiZhu, line 1: Don't commit db.sqlite3 now.

PR #126, opened 2023-03-24
Nyc open data

Hgersten-hash:
This pr adds a Django command to pull data from NYC Open Data and store it in the NYCharities model. The data set can be found here: [dataset](https://data.cityofnewyork.us/Social-Services/2019-Volunteers-Count-Report-Boroughs/yunp-vs8g)
There are also tests for the command in the tests folder


voluncheer/map/tests/test_nyc_data_com.py
```diff
@@ -0,0 +1,51 @@
...
+                "organization_type": "Community",
+                "street_address_mailing_address": "456 Test Street",
+                "city": "New York",
+                "state": "NY",
+            },
+        ]
+
+    @patch("map.management.commands.pull_nyc_charities.requests.get")
+    def test_command(self, mock_get):
+        mock_get.return_value.json.return_value = self.test_data
```
JinweiZhu, line 41: Where are you going to use mock_get?

JinweiZhu: What function is callign test_command and passing in a mock_get?

voluncheer/map/management/commands/pull_nyc_charities.py
```diff
@@ -0,0 +1,39 @@
...
+    def handle(self, *args, **options):
+        url = "https://data.cityofnewyork.us/resource/yunp-vs8g.json"
+
+        response = requests.get(url)
+
+        if response.status_code != 200:
+            raise CommandError("Could not pull data from NYC Open Data")
+
+        # Clear NYCharities table before inserting new data
+        NYCharities.objects.all().delete()
```
joeleu, line 22: Deleting all of the existing entries may not be supportable when using more durable storage. Since it will essentially change the primary key out from under us, if that key is referenced anywhere else that link will break. We should support updating existing entries and creating new ones. As far as removing old ones we can probably ignore that for now.

WDYT?

voluncheer/map/models.py
```diff
@@ -1,3 +1,18 @@
-# from django.db import models
+from django.db import models
+
 
 # Create your models here.
+class NYCharities(models.Model):
+    class Meta:
+        verbose_name_plural = "NYCharities"
```
joeleu, line 7: Are these all across NY or just in NYC? If the latter do you think we should have two "C"s in the name? NYCCharities?

voluncheer/map/models.py
```diff
@@ -1,3 +1,18 @@
...
+
+    name = models.CharField(max_length=255)
+    street = models.CharField(max_length=255)
+    city = models.CharField(max_length=255)
+    state = models.CharField(max_length=2)
+    type = models.CharField(max_length=255)
+    latitude = models.FloatField(default=0.0)
```
joeleu, line 14: WDYT about using the models.PointField for the latitude and longitude? The GeoDjango model API might be useful here.

https://docs.djangoproject.com/en/4.1/ref/contrib/gis/model-api/

voluncheer/map/tests/test_nyc_data_com.py
```diff
@@ -0,0 +1,51 @@
...
+            {
+                "organization_city_agency": "Test Charity 3",
+                "organization_type": "Community",
+                "street_address_mailing_address": "456 Test Street",
+                "city": "New York",
+                "state": "NY",
+            },
+        ]
+
+    @patch("map.management.commands.pull_nyc_charities.requests.get")
```
joeleu, line 39: Please prefer using the `mock.patch.object` method instead: https://docs.python.org/3/library/unittest.mock.html#patch-object

Using `patch` directly makes static dependency analysis impossible.

joeleu (changes requested):
We will need to coordinate how to populate this information in the AWS environments but we can do that in a follow up PR. Nice work.

voluncheer/map/management/commands/pull_nyc_charities.py
Hgersten-hash: Makes sense! I will change it to update existing entries and create new ones.

voluncheer/map/models.py
Hgersten-hash: Visually, I like the one C better, but if we are being very specific I can change the name.

voluncheer/map/management/commands/pull_nyc_charities.py
joeleu: Great, let me know when that's done and I can re-review.

voluncheer/map/models.py
joeleu: Nah, that's fine. Just wanted to make sure we were thinking about it.

voluncheer/map/models.py
dbleek: When working on the opportunity model, my initial research gave me the impression that GeoDjango was more than is necessary for our app, but I can't really point at any reason why. 

If we are only using the the longitude and latitude to plot a point on a map, it may be simpler to just use separate float or decimal fields for lon and lat as you have already. This would at the very least save us from importing GeoDjango models in addition to vanilla django.db models.

dbleek (approved):
Looks good to me, not much to add to what's been said.

voluncheer/map/models.py
dbleek: Btw I landed on using decimal fields for lon and lat based on this: https://stackoverflow.com/questions/30706799/which-model-field-to-use-in-django-to-store-longitude-and-latitude-values

The main takeaway seems to be that float, decimal and point are all viable options.

joeleu: Ack, thanks for the explanation Doug. I just wanted to be sure we were considering all the options. Decimal seems like the best option. Float is always a bit scary to me given its inherent inaccuracy.

voluncheer/map/management/commands/pull_nyc_charities.py
Hgersten-hash: Done! Now it doesn't delete the objects, but it uses the update_or_create method to find a charity with the provided info, and updates the defaults if found. I had to add a bin_num attribute because it was unique for each entry in the data set. This way, if a charity changes its location, we can update the new address and show this on the map.

voluncheer/map/models.py
Hgersten-hash: I saw that too @dbleek! Thanks for sending the link. I didn't think the GeoDjango was necessary since we are using Google Maps

voluncheer/map/tests/test_nyc_data_com.py
Hgersten-hash: Fixed. It now uses the patch.object.

voluncheer/db.sqlite3
JinweiZhu, line 1: Don't commit db sqlite3 now.

voluncheer/map/models.py
```diff
@@ -1,3 +1,19 @@
...
+class NYCharities(models.Model):
+    class Meta:
+        verbose_name_plural = "NYCharities"
+
+    name = models.CharField(max_length=255)
+    street = models.CharField(max_length=255)
+    city = models.CharField(max_length=255)
+    state = models.CharField(max_length=2)
+    type = models.CharField(max_length=255)
+    bin_num = models.PositiveBigIntegerField(default=0)
```
joeleu, line 14: I'm not sure I understand what the `bin_num` provides us that the ID of the object does not. Can you clarify it a bit? Also, what does the name mean?

joeleu (changes requested):

voluncheer/map/models.py
Hgersten-hash: There is a field in the data named bin; there isn't a description for it, but I think it is like a bank number. There is no ID number for the NYCharities data set - that's why I am using the combination of the organization name, organization type, and bin number as a way to check if a record in the data set already exists in our NYCharities model. If it does, then we can update the address if there are any changes.

voluncheer/db.sqlite3
Hgersten-hash: How can I remove the db sqlite from a commit?

voluncheer/map/models.py
Hgersten-hash: My concern here was that the address for an nycharity object would be updated if necessary since this is important for our map.

voluncheer/db.sqlite3
JinweiZhu: git rm -cached db.sqlite3

voluncheer/map/models.py
joeleu: I see. Can you provide a comment just above the `bim_num = ...` line explaining what it is then? When a field can't be easily identified (without prior knowledge) just by looking at the code, that's when comments come in handy. Better yet, if you can provide a link or reference to the source that would be ideal.

Hgersten-hash: By link/ reference do you mean a reference to the data set?

Hgersten-hash: I added the comments and the link to the data set.

Hgersten-hash: Let me know if it's what you had in mind

Hgersten-hash: I definitely committed it last night... and it shows the new comments on my branch. I'll commit again

PR #128, opened 2023-03-26
Profile url cleanup

dbleek:
Similar to my last attempt, but up-to-date with the current develop branch. This PR:

- Cleans up the project root urls.py by moving out urls specific to the profiles app to the correct namespace 
- Removes unnecessary regex anchors from profiles.urls.py
- Fixes time-zone warnings in opportunityboard tests


JinweiZhu (approved):
LGTM

voluncheer/voluncheer/urls.py
```diff
@@ -18,32 +18,16 @@
...
-    path(
-        "accounts/signup/volunteer/",
-        VolunteerSignUpView.as_view(),
-        name="volunteer_signup",
-    ),
-    # Unimplemented urls.
-    # path("chat/", include("chatroom.urls")),
-    path("opportunityboard/", include("opportunityboard.urls")),
+    path("admin/", admin.site.urls),
+    path("__debug__/", include("debug_toolbar.urls")),
```
joeleu, line 28: I think you'll want to only include this when `settings.DEBUG` is true. If not, what happens when `settings.DEBUG` is false and someone visits this url?

dbleek: So, if `settings.DEBUG` is false, `debug_toolbar ` is not added to installed apps. So if someone enters that url, it doesn't find anything because `debug_toolbar.urls `doesn't exist. So we get a standard 404 error.

Interestingly, we get a 404 for this url as well if debug mode is on, so this url must just be something the toolbar needs to render as opposed to an actual page.

joeleu: Hmm, that's interesting. Thanks for figuring it out. I assume we would still only want to include it when `settings.DEBUG` is true though. We don't know what we are risking by including the URL in a non-debug environment.

voluncheer/profiles/urls.py
```diff
@@ -1,14 +1,27 @@
...
 from profiles.views.home import home
+from profiles.views.organizations import OrganizationSignUpView
 from profiles.views.profile import ProfileView
 from profiles.views.profile import profile_update
+from profiles.views.volunteers import VolunteerSignUpView
 
 urlpatterns = [
     path("", home, name="home"),
-    re_path(r"^profile/$", ProfileView.as_view(), name="profile"),
-    re_path(r"^profile/update/$", profile_update, name="profile_update"),
```
EdenWuyifan, line 12: I added that because there's a domain issue when I update profile infos. Can you try testing our functions and see if they go well?

EdenWuyifan (changes requested):

voluncheer/profiles/urls.py
dbleek: I tested both volunteer and organization updating and both worked well.

My understanding of `^profile/$` is that it would match any strings that both start and end with "profile/", so wouldn't that just be "profile/"? Or "profile/profile/profile/"?

voluncheer/voluncheer/urls.py
dbleek: Sounds good, I made the requested change in urls.py

voluncheer/profiles/urls.py
joeleu: The `^profile/$` regular expression basically says the the path must be exactly `profile/`, one a single occurrence of the content. That said, I would not expect the regular expression to be needed here.

PR #130, opened 2023-03-26
Automate categories database

JinweiZhu:
Hi Joe, so I saved Doug's category data as a json file in opportunityboard/fixtures/category_data.json. The ./manage.py loaddata command would populate this data in the database right after migration. This would take care of local and CI. But I am not sure if AWS automatically do the same. Are we not running migrations on AWS? If so, how are we going to upload the json script data into the RDS? I don't currently have the permission to play around with the AWS RDS UI.


.github/actions/prepare/action.yml
```diff
@@ -23,3 +23,4 @@ runs:
       run: |
         python manage.py makemigrations
         python manage.py migrate
+        python manage.py loaddata opportunityboard/fixtures/category_data.json
```
joeleu, line 26: It's not clear to me why we need to inject this data during CI. Tests should be self contained and therefore they should add any data they rely on being in the database at test execution time.

voluncheer/opportunityboard/fixtures/category_data.json
```diff
@@ -0,0 +1,309 @@
+[
+    {
+      "model":"opportunityboard.category",
+      "pk":1,
```
joeleu, line 4: I would expect the PK to be generated when data is added to the database, not something we should synthetically create. Can you explain your thinking around including it here?

voluncheer/opportunityboard/fixtures/category_data.json
```diff
@@ -0,0 +1,309 @@
...
+      }
+    },
+    {
+      "model": "opportunityboard.subsubcategory",
+      "pk": 19,
+      "fields": {
+        "name": "Schools",
+        "parent_id": 6
+        }
+        },
```
joeleu, line 256: The format of this file went a bit awry here, please keep consistent.

voluncheer/opportunityboard/fixtures/category_data.json
```diff
@@ -0,0 +1,309 @@
...
+        {
+        "model": "opportunityboard.subsubcategory",
+        "pk": 25,
+        "fields": {
+        "name": "Soup Kitchen",
+        "parent_id": 7
+        }
+        }
+
```
joeleu, line 305: Couple of extra newlines here that aren't necessary.

voluncheer/opportunityboard/migrations/0003_alter_opportunity_photo.py
joeleu, line 1: Should we be deleting migrations files? What does doing so imply?

voluncheer/profiles/migrations/0002_auto_20230325_2115.py
joeleu, line 1: Same question as above regarding deleting migrations files.

voluncheer/profiles/urls.py
```diff
@@ -1,14 +1,27 @@
...
+from profiles.views.organizations import OrganizationSignUpView
 from profiles.views.profile import ProfileView
 from profiles.views.profile import profile_update
+from profiles.views.volunteers import VolunteerSignUpView
 
 urlpatterns = [
     path("", home, name="home"),
-    re_path(r"^profile/$", ProfileView.as_view(), name="profile"),
-    re_path(r"^profile/update/$", profile_update, name="profile_update"),
+    path("profile/", ProfileView.as_view(), name="profile"),
```
joeleu, line 13: These are changes already staged in Doug's PR.

voluncheer/voluncheer/urls.py
```diff
@@ -18,32 +18,16 @@
...
-    ),
-    path(
-        "accounts/signup/volunteer/",
-        VolunteerSignUpView.as_view(),
-        name="volunteer_signup",
-    ),
-    # Unimplemented urls.
-    # path("chat/", include("chatroom.urls")),
-    path("opportunityboard/", include("opportunityboard.urls")),
+    path("admin/", admin.site.urls),
```
joeleu, line 27: Same here. These are already staged in another PR.

joeleu (changes requested):

voluncheer/opportunityboard/fixtures/category_data.json
JinweiZhu: I just copied Doug's hardcoded categories and convert them into json file. I guess it is not necessary but it does make it more human readable.

voluncheer/opportunityboard/migrations/0003_alter_opportunity_photo.py
JinweiZhu: Sometimes, there are irresolvable migration conflicts due to git, and the easiest solution is to just remove them.

.github/actions/prepare/action.yml
JinweiZhu: This is not exactly testing. This is let Django load some important hard-coded categories into the database so users can actually see them in the UI.

voluncheer/opportunityboard/fixtures/category_data.json
joeleu: My point is those values may already exist in a database you are going to attempt to write them to. It isn't safe to assume those values are not already used.

.github/actions/prepare/action.yml
joeleu: Right and the database being used during the GitHub actions workflows is _only_ for testing, thus these are not required.

voluncheer/opportunityboard/migrations/0003_alter_opportunity_photo.py
joeleu: How does that affect a non-sqlite database? How can we be sure it won't affect our production databases?

voluncheer/opportunityboard/fixtures/category_data.json
JinweiZhu: Removed.

voluncheer/opportunityboard/migrations/0003_alter_opportunity_photo.py
JinweiZhu: Okay added back history in new PR>

voluncheer/profiles/urls.py
JinweiZhu: This is not in new PR.

PR #131, opened 2023-03-26
Recurrences

dbleek:
This PR:

- Adds `is_recurring`, `recurrence` and `end_date `fields to the Opportunity model
- Includes these fields on the post/update opportunity templates, and disables `recurrence` and `end_date` when  `is_recurring=False`
- Raises an error is `is_recurring=True` and  `recurrence=None` during form validation
- Streamlines the Opportunity save for less repetition/easier customization


voluncheer/opportunityboard/views/postanopportunity.py
```diff
@@ -47,14 +51,18 @@ def update_an_opportunity(request, opportunity_id):
...
-        form = PostAnOpportunityForm(request.POST, request.FILES)
+        form = PostAnOpportunityForm(request.POST, request.FILES, instance=opportunity_to_update)
         if form.is_valid():
             if "delete" in request.POST:
                 form.delete(opportunity_id)
             else:
-                form.update(opportunity_id)
+                form.save()
         else:
-            print(form.errors.as_data())
```
JinweiZhu, line 57: We shouldn't render if the form is invalid. We should throw an error right?

voluncheer/opportunityboard/test_forms.py
```diff
@@ -82,8 +84,28 @@ def test_update(self):
...
+            "end": "12:00:00",
+            "address_1": "New York, NY",
+            "address_2": "Mean st",
+            "is_published": True,
+            "is_recurring": True,
+            "recurrence": None,  # Recurrence set to none despite is_recurring = True
+        }
+
+        form = PostAnOpportunityForm(data=data)
+        self.assertFalse(form.is_valid())
```
JinweiZhu, line 111: We should have at least one test that make sure the new fields added above are set to the desired value, no?

voluncheer/opportunityboard/models.py
```diff
@@ -80,6 +85,9 @@ class Meta:
     description = models.TextField()
     date = models.DateTimeField()
     end = models.TimeField()
+    is_recurring = models.BooleanField(default=False)
+    recurrence = models.CharField(max_length=6, choices=FREQUENCIES, null=True, blank=True)
```
JinweiZhu, line 89: Are you reusing FREQUENCIES anywhere else?

voluncheer/opportunityboard/views/postanopportunity.py
```diff
@@ -26,7 +26,11 @@ def post_an_opportunity(request):
         if form.is_valid():
             form.save()
         else:
-            print(form.errors.as_data())
```
JinweiZhu, line 29: If the request is a post and form is invalid, then it shouldn't render.

voluncheer/opportunityboard/test_forms.py
dbleek: Added fields to test_models.py.

voluncheer/opportunityboard/models.py
dbleek: Moved `FREQUENCIES `under Opportunity model definition

voluncheer/opportunityboard/views/postanopportunity.py
dbleek: If we don't render, the user will not see the error message and will not know why the opportunity failed to be created.

voluncheer/opportunityboard/views/postanopportunity.py
dbleek: See above

PR #137, opened 2023-03-27
fixed end_date bug

dbleek:
This is a one-line change to address a bug with `end_date` field in the `PostAnOpportunityForm`, which corrects the format in which the date is submitted.

Not entirely sure why the DateField is complaining about the date in a d-m-y format, but changing it to y-m-d resolved the issue. This doesn't affect how it displays to the user/on the form.


Hgersten-hash (approved):
I think it has something to do with overriding Django’s default expectation for a DateField object.

PR #141, opened 2023-03-27
Adding command to deployment

Hgersten-hash:
This pr adds the pull_nyc_charities command to the django.config file so that the command is executed at deployment.


voluncheer/opportunityboard/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 3.2.8 on 2023-03-27 01:10
+# Generated by Django 4.0 on 2023-03-27 02:26
```
JinweiZhu, line 1: Why are we switching from django 4.0 instead of 3.2.8?

voluncheer/map/admin.py
```diff
@@ -1,3 +1,6 @@
-# from django.contrib import admin
+from django.contrib import admin
+
+from .models import NYCharities
```
joeleu, line 3: Please use absolute paths for imports:

```suggestion
from map.models import NYCharities
```

joeleu (changes requested):
I think much of this is already in develop so you may need to rebase these changes on top of develop to get a more accurate diff.

PR #151, opened 2023-03-28
change filter to OOP, add map icon and search placeholder

EdenWuyifan:


voluncheer/opportunityboard/views/search.py
```diff
@@ -9,12 +9,73 @@
...
 def filter_search(request):
+    filter = Filter()
     if request.method == "POST":
         filter = parse_search_filter(request.POST)
         logger = logging.getLogger(__name__)
-        logger.error(filter)
-        opportunity_lists = search_by_filters(filter)
+        logger.error(filter.gen_dict())
+        opportunity_lists = filter.search()
     else:
```
dbleek, line 79: This is essentially a repeat of the opportunityboard view right? Could we use `return render HttpRedirectResponse(reverse(opportunityboard))` instead? Or is there some way we can avoid repetition?

Also, since we are not modifying/inserting data, can we use GET instead of POST? In that case, we change the form to GET and then there is only one branch for `search()`? Just an idea.

voluncheer/opportunityboard/views/search.py
```diff
@@ -9,12 +9,73 @@
...
+
+
 def filter_search(request):
+    filter = Filter()
     if request.method == "POST":
         filter = parse_search_filter(request.POST)
         logger = logging.getLogger(__name__)
-        logger.error(filter)
-        opportunity_lists = search_by_filters(filter)
+        logger.error(filter.gen_dict())
```
dbleek, line 77: Do we need the logger in production?

voluncheer/opportunityboard/tests_views.py
```diff
@@ -91,33 +91,15 @@ def test_search(self):
         self.assertEqual(response.status_code, 200)
 
     def test_search_by_filters(self):
```
dbleek, line 93: Please add a docstring here to explain the test.

voluncheer/opportunityboard/tests_views.py
dbleek, line 1: This could be broken down into 4 smaller tests for clarity--test_parse_search_filter, test_search_by_category, test_search_by_subcategory, etc..

dbleek (changes requested):

voluncheer/opportunityboard/views/search.py
EdenWuyifan: Yeah we can keep those for now

dbleek: What does the logger do?

EdenWuyifan: It's just a graceful way of log print. Since we are expanding for more filters, it is nice to put a log here in order to find out if the filter works

dbleek:
Once you add the doc string for the test, we should be all set

voluncheer/opportunityboard/tests_views.py
EdenWuyifan: Done

dbleek (commented):
All set, thanks for the changes Eden!

dbleek (approved):

PR #155, opened 2023-03-29
Merge code to seed NYCharities into main

joeleu:


voluncheer/.ebextensions/django.config
```diff
@@ -2,6 +2,9 @@ container_commands:
   01_migrate:
     command: "source /var/app/venv/*/bin/activate && python3 manage.py migrate"
     leader_only: true
+  02_pull_data:
+    command: "source /var/app/venv/*/bin/activate && python3 manage.py pull_nyc_charities"
```
JinweiZhu, line 6: If this should only happen once, then it should not be here. It should be entered on the console of AWS.

JinweiZhu (approved):

dbleek (approved):
Looks good to me

PR #158, opened 2023-03-31
Chatroom

JinweiZhu:
I am following this tutorial to build a [chatroom.](https://testdriven.io/blog/django-channels/)
I already have the chat_homepage and room ready. I stopped before the Authentication section. I plan to continue after the current UI is updated. Right now, the chat is real time, doesn't restrict attendees, and doesn't persist in the database after Redis shut down.


JinweiZhu:
Correct link added. Sorry.

EdenWuyifan:
Is there a way I can test the functionalities?

.github/actions/prepare/action.yml
```diff
@@ -23,3 +23,8 @@ runs:
       run: |
         python manage.py makemigrations
         python manage.py migrate
+    - name: spin up redis with docker
+      shell: bash
+      working-directory: voluncheer
+      run: |
+        docker run -p 6379:6379 -d redis:5
```
joeleu, line 30: GitHub actions has a service container for Redis, we should use that instead: https://docs.github.com/en/actions/using-containerized-services/creating-redis-service-containers

voluncheer/chatroom/consumers.py
```diff
@@ -0,0 +1,50 @@
+import json
+
+from asgiref.sync import async_to_sync
+from channels.generic.websocket import WebsocketConsumer
+
+from .models import Room
```
joeleu, line 6:
```suggestion
from chatroom.models import Room
```

joeleu: Please prefer absolute imports over relative.

voluncheer/chatroom/views.py
```diff
@@ -0,0 +1,30 @@
+from django.contrib.auth.views import redirect_to_login
+from django.shortcuts import render
+
+from chatroom.models import Room
+
+
+# To be changed after UI is designed.
+def chat_homepage_view(request):
+    if not request.user.is_authenticated:
```
joeleu, line 9: There is a handy decorator that does this for you: https://docs.djangoproject.com/en/4.1/topics/auth/default/#the-login-required-decorator

voluncheer/requirements.txt
```diff
@@ -32,4 +32,5 @@ six==1.16.0
 sqlparse==0.4.3
 urllib3==1.26.14
 verify-email==2.4.3
-
+channels==3.0.4
```
joeleu, line 35: Please sort alphabetically

voluncheer/static/admin/js/chat_homepage.js
```diff
@@ -0,0 +1,23 @@
+console.log("Sanity check from chat_homepage.js.");
```
joeleu, line 1: Seems this got left in from debugging during development.

voluncheer/static/admin/js/room.js
```diff
@@ -0,0 +1,84 @@
+console.log("Sanity check from room.js.");
```
joeleu, line 1: Seems this got left in from debugging during development.

voluncheer/templates/chatroom/chat_homepage.html
```diff
@@ -0,0 +1,41 @@
+<!-- placeholder chat_homepage file -->
+
+{% load static %}
+
+<!DOCTYPE html>
```
joeleu, line 5: This should extend the base html template. See the profile page for an example.

voluncheer/templates/chatroom/room.html
```diff
@@ -0,0 +1,49 @@
+<!-- placeholder room file -->
+
+{% load static %}
+
+<!DOCTYPE html>
```
joeleu, line 5: Same comment as the chat_homepage.html file.

voluncheer/voluncheer/asgi.py
```diff
@@ -6,11 +6,18 @@
...
 """
-
 import os
 
+from channels.routing import ProtocolTypeRouter
+from channels.routing import URLRouter
 from django.core.asgi import get_asgi_application
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "voluncheer.settings")
+import chatroom.routing
```
joeleu, line 15:
```suggestion
from chatroom import routing
```

voluncheer/voluncheer/settings.py
```diff
@@ -91,8 +94,16 @@
...
 
 WSGI_APPLICATION = "voluncheer.wsgi.application"
+ASGI_APPLICATION = "voluncheer.asgi.application"
 
-
+CHANNEL_LAYERS = {
+    "default": {
+        "BACKEND": "channels_redis.core.RedisChannelLayer",
+        "CONFIG": {
+            "hosts": [("127.0.0.1", 6379)],
```
joeleu, line 103: What is the plan for AWS? Using a local container will not work especially when there is more than one instance running.

joeleu (changes requested):

voluncheer/voluncheer/settings.py
joeleu: I suggest looking into https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/ElastiCache-Getting-Started-Tutorials-Python.html for AWS integration.

voluncheer/templates/chatroom/chat_homepage.html
JinweiZhu: I tried it and it interferes with the demo interface.

voluncheer/templates/chatroom/room.html
JinweiZhu: Same, it interferes with the demo interface. So I suggest we make the change later.

.github/actions/prepare/action.yml
JinweiZhu: Right now it doesn't interfere with CI. Maybe I can work on this next week? I haven't built the test suite yet due to time-constraint.

joeleu: We aren't rushing this in so let's get this right.

voluncheer/static/admin/js/chat_homepage.js
JinweiZhu: I think this log really helps with debugging in production for JS related issues. I prefer to leave it in.

.github/actions/prepare/action.yml
JinweiZhu: Do you think I would need to test for chatroom in CI though? If so, what kind of test should that be? I am already testing for chatroom model and view access, and those doesn't seem to require redis.

joeleu: If the testing works without a redis provider we shouldn't need this at all.

voluncheer/static/admin/js/chat_homepage.js
joeleu: How so? What is this log line signaling to you on production? You can get this information from the Chrome debugging tools.

voluncheer/templates/chatroom/chat_homepage.html
joeleu: That is disappointing, what kind of issues are you experiencing? I can try it myself but not until later this week.

JinweiZhu: The reason being that I copied the html file from an app online. So it doesn't currently integrate with our other UI design. I think Eden will figure it out afterward.

JinweiZhu: In other words, this will be done but just not in this PR or story.

EdenWuyifan: I'll handle that, no need to worry

voluncheer/templates/chatroom/room.html
EdenWuyifan: Same here, will get it done as soon as this is passed to the frontend

.github/actions/prepare/action.yml
joeleu: Please remove if we are indeed not using this.

voluncheer/chatroom/views.py
```diff
@@ -0,0 +1,29 @@
...
+        "chatroom/chat_homepage.html",
+        {
+            "rooms": Room.objects.all(),
+        },
+    )
+
+
+@login_required
+def room_view(request, room_name):
+    room, created = Room.objects.get_or_create(name=room_name)
```
joeleu, line 22: Since created is not used you can just discard it by assigning to an `_`:

```suggestion
room, _ = Room.objects.get_or_create(name=room_name)
```

voluncheer/static/admin/js/chat_homepage.js
```diff
@@ -0,0 +1,23 @@
+
```
joeleu, line 1: Two superfluous blank lines, please remove.

voluncheer/static/admin/js/room.js
```diff
@@ -0,0 +1,84 @@
+
```
joeleu, line 1: Same here, please remove.

voluncheer/voluncheer/settings.py
```diff
@@ -91,6 +94,39 @@
 ]
 
 WSGI_APPLICATION = "voluncheer.wsgi.application"
+ASGI_APPLICATION = "voluncheer.asgi.application"
+
+redis_prod_url = "redis://chatroom-redis-cluster-prod.byxkua.ng.0001.use1.cache.amazonaws.com:6379"
```
joeleu, line 99: For environment specific variables, set them in the environment using variables and read them in here. See the various other cases in here for examples.

voluncheer/voluncheer/settings.py
```diff
@@ -166,7 +202,7 @@
 # https://docs.djangoproject.com/en/4.1/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
-
+ALLOW_ANONYMOUS = False
```
joeleu, line 205: What does this do?

joeleu (changes requested):

voluncheer/voluncheer/settings.py
```diff
@@ -91,6 +94,40 @@
 ]
 
 WSGI_APPLICATION = "voluncheer.wsgi.application"
+ASGI_APPLICATION = "voluncheer.asgi.application"
+
+
+REDIS_PROD_MAIN_PORT = os.getenv("REDIS_PROD_MAIN_PORT")
```
joeleu, line 100:
```suggestion
REDIS_MAIN_PORT = os.getenv("REDIC_MAIN_PORT")
```

joeleu: Then delete the prod/dev environment branches below in favor of a single branch when on aws. The don't have any difference other that the address/port.

joeleu (changes requested):

voluncheer/voluncheer/settings.py
JinweiZhu: Actually, they run on two different redis servers on the backend. I think there are good reasons to set up 2 servers so downing the develop doesn't result in the downing of production. Is it too expensive?

joeleu: But they are unique to the environment they connect to, right? The prod one connects to main and the dev one connects to develop I assume.

voluncheer/voluncheer/settings.py
```diff
@@ -91,6 +94,29 @@
 ]
 
 WSGI_APPLICATION = "voluncheer.wsgi.application"
+ASGI_APPLICATION = "voluncheer.asgi.application"
+
+
+REDIS_CHATROOM_PORT = os.getenv("REDIS_CHATROOM_PORT")
+if environment.is_production or environment.is_development:
```
joeleu, line 101: We have the `environment.is_aws` which obfuscates away the specific environment.

https://github.com/gcivil-nyu-org/INET-Monday-Spring2023-Team-2/blob/develop/voluncheer/voluncheer/environment.py#L24-L26

joeleu (approved):

PR #159, opened 2023-04-01
add pagination to opportunityboard, update testcases

EdenWuyifan:


voluncheer/opportunityboard/tests_views.py
```diff
@@ -77,7 +77,7 @@ def test_update_an_opportunity_page_loads(self):
     def test_search(self):
         """Tests update_an_opportunity page loads"""
         rf = RequestFactory()
-        post_request = rf.post(
```
JinweiZhu, line 80: WHy is update_an_opportunity get not post?

voluncheer/opportunityboard/views/search.py
```diff
@@ -69,26 +69,6 @@ def search(self):
         return filtered_opportunity
 
 
-def filter_search(request):
```
JinweiZhu, line 72: Why are we removing filter search?

voluncheer/opportunityboard/tests_views.py
EdenWuyifan: Since for search we only need to read data, not write, so a get is enough

voluncheer/opportunityboard/views/search.py
EdenWuyifan: For pagination, we can merge the search func and the display. Since other cases there's no filter submit. Reading filter won't return anything.

JinweiZhu (approved):
LGTM

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -1,27 +1,42 @@
...
+OPPORTUNITY_PER_PAGE = 5
 
 
-def opportunityboard(request):
+def opportunityboard(request, page_number):
     """View for Opportunity Board"""
-    opportunity_lists = Opportunity.objects.order_by("-pubdate"[:20])
+    filter = parse_search_filter(request.GET)
+    logger = logging.getLogger(__name__)
+    logger.error(filter.gen_dict())
```
joeleu, line 21: We should not log errors during normal, successful operations. Polluting the error logs with non-error messages makes debugging more difficult.

voluncheer/templates/navbar.html
```diff
@@ -10,7 +10,7 @@
       </a>
     </li>
     <li>
-      <a href="{% url 'opportunityboard' %}">
+      <a href="{% url 'opportunityboard' 1%}">
```
joeleu, line 13: Does this `1` actually do anything?

joeleu (changes requested):

voluncheer/templates/navbar.html
EdenWuyifan: It's for default page number

JinweiZhu (approved):

PR #160, opened 2023-04-01
Filter by duration

dbleek:
This PR implements the following:

- When creating/updating an opportunity, validates that the start time is before the end time
- Filter by duration search: users can select the following options:  2 hours or less, 4 hours or less, Full-day (8 hours or less)


voluncheer/opportunityboard/test_forms.py
```diff
@@ -54,7 +54,7 @@ def test_validation(self):
             "description": "Let's surfing",
             "staffing": "1",
             "date": timezone.now(),
```
joeleu, line 56: I see this isn't in this PR but I would be remise to leave it alone. Generally you should never use a time.now() like function in a test, it makes the test non-deterministic. Rather you should create a "fake" date and time such that it is always the same.

voluncheer/opportunityboard/tests_views.py
```diff
@@ -104,3 +107,83 @@ def test_search_by_categories(self):
         filters.subcategory = None
         filters.subsubcategory = self.subsubcategory
         self.assertEqual(filters.search().count(), 1)
+
+    def test_search_by_duration_lte2hrs(self):
```
joeleu, line 111: These tests are quite difficult to read. It would be a bit easier if the event started at 0:00 and then you set the durations based on incrementing from there. It's not realistic from and actual opportunity point of view but the logic will still apply and when someone reads it they will know that the duration is 1.5 hours when the end time is 1:30. Thoughts?

The tests seem logically sound, just hard to read.

You might also want to look into creating "subTest"s within each test function to differentiate and annotate which durations you are testing and which results you are expecting: https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest

Something like: `with self.subTest("one_hour_exists"):` etc.

voluncheer/opportunityboard/views/search.py
```diff
@@ -65,10 +66,32 @@ def search(self):
...
+        if self.duration == "2-4 Hours":
+            filtered_opportunity = filter_by_duration(filtered_opportunity, 2, 4)
+        if self.duration == "4-8 Hours":
+            filtered_opportunity = filter_by_duration(filtered_opportunity, 4, 8)
+        if self.duration == "8+ Hours":
+            filtered_opportunity = filter_by_duration(filtered_opportunity, 8, 24)
         return filtered_opportunity
 
 
+def filter_by_duration(queryset, min, max):
```
joeleu, line 80: Did you considered using just a maximum instead of a range? I'm wondering what caused you to go for a range rather than just a maximum.

voluncheer/opportunityboard/views/search.py
```diff
@@ -65,10 +66,32 @@ def search(self):
...
+    """
+    Takes Opportunity queryset, as well as a min and a max number of hours
+    Returns Opportunity queryset with durations only between min and max hours
+    """
+    min_duration = dt.timedelta(hours=min)
+    max_duration = dt.timedelta(hours=max)
+    for opportunity in queryset:
```
joeleu, line 87: An alternative would be to accumulate the opportunities that match the criteria via something like list.append instead of excluding one's that do not match. It's mostly a personal choice but I generally find inclusion easier to read that exclusion. Up to you though.

voluncheer/opportunityboard/views/search.py
```diff
@@ -65,10 +66,32 @@ def search(self):
...
             filtered_opportunity = filtered_opportunity.filter(subsubcategory=self.subsubcategory)
-
+        if self.duration == "0-2 Hours":
+            filtered_opportunity = filter_by_duration(filtered_opportunity, 0, 2)
+        if self.duration == "2-4 Hours":
+            filtered_opportunity = filter_by_duration(filtered_opportunity, 2, 4)
+        if self.duration == "4-8 Hours":
+            filtered_opportunity = filter_by_duration(filtered_opportunity, 4, 8)
+        if self.duration == "8+ Hours":
+            filtered_opportunity = filter_by_duration(filtered_opportunity, 8, 24)
```
joeleu, line 76: If this maxes out at 24 hours it's not really 8+ but rather 8-24, right?

joeleu (changes requested):
Mostly looks good, nice work.

voluncheer/opportunityboard/test_forms.py
dbleek: The `setUp() `test cases for opportunity board were a mess, so I created a helper function in unittest_setup.py that could be imported to make the test opportunities more consistent. Now all tests avoid using `timezone()` to create the start datetime.

voluncheer/opportunityboard/views/search.py
dbleek: Looking over the acceptance criteria, I think your suggestion is closer to what was specified, so I changed the function accordingly. Assuming "Full-day" means 8 hours, but curious what Hanna thinks.

voluncheer/opportunityboard/views/search.py
dbleek: Due to the changes above (searching for durations less than equal to max) this is no longer an issue. Not sure if we will want allow for opportunities greater than 8 hours and how we should search for those, but will discuss with Hanna and address if needed.

voluncheer/opportunityboard/views/search.py
dbleek: My thought process was that we already have a queryset filtered by category without regard to duration.  It feels more intuitive to me to further hone down the existing queryset by removing opportunities that are too long, rather than creating a new queryset that appends items from the existing one, at least given the option of Django's `filter ` vs. `exclude`.

voluncheer/opportunityboard/tests_views.py
dbleek: Followed your advice on the 0:00 start time for the test opportunity. Please let me know what you think.

voluncheer/opportunityboard/unittest_setup.py
```diff
@@ -0,0 +1,63 @@
...
+from opportunityboard.models import Category
+from opportunityboard.models import Opportunity
+from opportunityboard.models import Subcategory
+from opportunityboard.models import Subsubcategory
+from profiles.models import Organization
+from profiles.models import User
+from profiles.models import UserType
+
+
+def setup_oppboard_tests(test_case):
```
joeleu, line 12: Good idea, moving this to a shared location. If I can make one suggestion though, a common pattern is to declare a new `TestCase` class that extends the `setUp` method. For example:

```
from django import test

class TestCase(test.TestCase):

    def setUp(self):
        super().setUp()
        # perform additional setup for this shared test case...
```

In the files that need this additional setup:

```
from opportunityboard.test import TestCase

class OpportunityTest(TestCase):

    # Optional additional setup...
    def setUp(self):
        super().setUp()

# Do tests...
```

WDYT?

voluncheer/opportunityboard/tests_views.py
joeleu: Might need to push a commit? I don't see any difference.

voluncheer/opportunityboard/unittest_setup.py
dbleek: Awesome, thanks for the tip. I changed the tests accordingly.

voluncheer/opportunityboard/tests_views.py
dbleek: Make sure you are viewing the changes from all commits. I noticed the same experience you had, the diffs didn't contain my latest updates. Might be a bug in GitHub

joeleu: Now I see it. Looks good

joeleu (approved):

voluncheer/opportunityboard/views/search.py
Hgersten-hash: Originally, I tried to think about typical time slots people would consider when signing up for an opportunity, and I first thought of a small, part-time, and full-time opportunity. I assumed that an opportunity would not typically exceed 8 hours.

Hgersten-hash: For now, I think that we can assume an opportunity 6 hours or above is considered full time, and would therefore show up in a search for full time opportunities

PR #161, opened 2023-04-02
Add the NYCharities to the map with details upon highlight.

joeleu:
This PR adds the NY Charities data to the map collapsed into an icon by default. Upon hovering over an icon some of the basic details for the organization are presented to the.

Example screenshot:

Screenshot 2023-04-01 at 21 12 41


voluncheer/static/css/map.css
```diff
@@ -40,5 +44,163 @@ body {
 }
 
 .text-center {
-    text-align: center!important;
+    text-align: center !important;
+}
+
+.org {
+    background-color: #4285F4;
```
Hgersten-hash, line 51: Can we add a border around the markers so that it is easier to differentiate when they are close together?

Hgersten-hash (changes requested):
LGTM! Can we just add a border around the markers so that it is easier to differentiate when they are close together?

JinweiZhu (approved):
Good job.

EdenWuyifan (approved):
LGTM

PR #172, opened 2023-04-02
Volunteer Signup

EdenWuyifan:


voluncheer/templates/opportunityboard/updateanopportunity.html
joeleu, line 1: These edits don't seem necessary?

voluncheer/opportunityboard/models.py
```diff
@@ -67,6 +68,7 @@ class Opportunity(models.Model):
         longitude: used for mapping the opportunity. *allowed to be blank for now
         latitude: used for mapping the opportunity. *allowed to be blank for now
         staffing: the requested number of volunteers for the opportunity.
+        volunteer: the list of registered users to the event.
```
joeleu, line 71: If it is more than one object we should be using the plural for the attribute. e.g. `volunteers`

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -48,3 +53,32 @@ def category_dict_gen():
             cate_dict[subcategory.name] = subsublist
         cate_output_dict[category.name] = cate_dict
     return cate_output_dict
+
+
+def signup_volunteer(request, opportunity_id):
+    volunteer = get_object_or_404(Volunteer, pk=request.user.pk)
+    opportunity = Opportunity.objects.get(pk=opportunity_id)
+    if opportunity.staffing > 0 and volunteer not in opportunity.volunteer.all():
```
joeleu, line 61: We need a way to not only identify if an opportunity requires staffing but also that the opportunity needs more staff. We need to track how many people have signed up and whether or not there are more staffing slots available. I recommend creating a method on the model that performs all this logic.

We need to keep the staffing number at the value set by the user so they can edit it later if they need to. We need an additional field to track the number of available slots, etc.

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -48,3 +53,32 @@ def category_dict_gen():
...
+        opportunity.staffing += 1
+        opportunity.save()
+    return HttpResponseRedirect(reverse("opportunityboard"))
+
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_volunteer_from_user(user, attr):
```
joeleu, line 82: Is this actually used anywhere?

joeleu (changes requested):

voluncheer/opportunityboard/views/opportunityboard.py
EdenWuyifan: Removed

voluncheer/opportunityboard/views/opportunityboard.py
EdenWuyifan: I do think we can calculate those from the staffing number and the opportunity.volunteer array.

EdenWuyifan: But yeah,  "We need to keep the staffing number at the value set by the user so they can edit it later if they need to. We need an additional field to track the number of available slots, etc." This makes sense, we might need to keep somewhere the original staffing

voluncheer/opportunityboard/models.py
EdenWuyifan: Changed

voluncheer/opportunityboard/tests_views.py
```diff
@@ -104,3 +107,76 @@ def test_search_by_categories(self):
         filters.subcategory = None
         filters.subsubcategory = self.subsubcategory
         self.assertEqual(filters.search().count(), 1)
+
+
+class VolunteerSignUpView(TestCase):
+    """This is the test case for Volunteer signup and other volunteer centric behaviors"""
+
+    def setUp(self):
```
joeleu, line 115: Be sure to call `super().setUp()` when extending the `setUp` method.

joeleu (changes requested):

voluncheer/opportunityboard/views/opportunityboard.py
joeleu: We need to make sure this doesn't get dropped either by creating an issue to track it or at least adding a TODO comment.

joeleu (approved):
Approval, pending some kind of tracking for staffing.

PR #174, opened 2023-04-03
Sprint Week 10

Hgersten-hash:
This pr merges the new features added for sprint week 10. They include map, pagination, filter by duration, and volunteer sign up.


dbleek (approved):
All looks good to me

PR #179, opened 2023-04-05
Feature eden volunteerhistory

EdenWuyifan:
**1. Fix bugs noted by Hanna:**
  - If you edit the About Me/Us sections, you must check box next to photo if you want to keep it after editing. Otherwise, it will disappear
  - When editing the About Me section on volunteer profile, you must add badges. Just write None for now
  - The profile photo disappears when you go to My Voluncheers 

**2. Add a list of Volunteers who signed up for the event.**
**3. Add attendance sheet for organizations to track if an volunteer showed up by:**
  - Add frontend checkbox, organization users check and updated with different attend status.
  - Backend add a new model called ArchivedOpportunity, storing the useful information of attended opportunities, which is useful for tracking hours.

image


voluncheer/opportunityboard/models.py
```diff
@@ -99,7 +99,10 @@ class Meta:
     staffing = models.PositiveIntegerField(null=True, blank=True)
     is_published = models.BooleanField(default=False)
     photo = models.ImageField(upload_to=_opportunity_photo_path, blank=True, null=True)
-    volunteers = models.ManyToManyField(Volunteer, blank=True)
+    volunteers = models.ManyToManyField(Volunteer, blank=True, related_name="volunteers")
+    attanded_volunteers = models.ManyToManyField(
+        Volunteer, blank=True, related_name="attanded_volunteers"
```
joeleu, line 104:
```suggestion
        Volunteer, blank=True, related_name="attended_volunteers"
```

voluncheer/opportunityboard/models.py
```diff
@@ -114,3 +117,47 @@ def duration(self):
         duration = end_datetime - start_datetime
         duration_seconds = dt.timedelta(seconds=duration.seconds)  # remove date
         return duration_seconds
+
+
+class ArchivedOpportunity(models.Model):
```
joeleu, line 122: Duplicating all of these fields would not be my choice for strategy. You could create a class that inherits from Opportunity above or, better yet, add the necessary fields to support archiving to the Opportunity class above.

voluncheer/profiles/forms/volunteers.py
```diff
@@ -51,15 +51,14 @@ def save(self, commit=True):
 
 class VolunteerChangeForm(UserChangeForm):
     password = None
-    photo = forms.ImageField(required=False)
+    # photo = forms.ImageField(required=False)
```
joeleu, line 54: If the field isn't needed, delete it don't comment it out. Comments like this just bloat and complicate the codebase.

voluncheer/profiles/tests/test_views.py
```diff
@@ -74,3 +77,26 @@ def test_signup_form_exists_at_desired_location(self):
...
+        get_request = rf.get(
+            f"/attandance/{self.opp.pk}",
+            {"volunteer-attanded": [str(self.vol.pk)]},
+        )
+        get_request.user = self.org
+        confirm_attandance(get_request, self.opp.pk)
+        self.assertEqual(self.opp.attanded_volunteers.count(), 0)
+        self.opp.volunteers.add(self.vol)
+        self.opp.refresh_from_db()
+        confirm_attandance(get_request, self.opp.pk)
```
joeleu, line 97: There are three different assertions going on here based on different actions. Please provide some kind of separation to make it clear where a particular test/assertion is occurring. One good option is the use of subTest: https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

Another thing to keep in mind when writing tests is the arrange, act, assert format. There are many posts on this topic, here is one as an example: https://learn.microsoft.com/en-us/visualstudio/test/unit-test-basics?view=vs-2022

voluncheer/profiles/views/profile.py
```diff
@@ -47,16 +49,17 @@ def get_context_data(self, **kwargs):
...
-            badge_list = kwargs["volunteer"].badges.split(",")
-            for badge in badge_list:
-                try:
-                    badge_urls.append(Volunteer.BADGES[badge])
-                except KeyError:
-                    pass
-            kwargs["badge_urls"] = badge_urls
             volunteer_profile = Volunteer.objects.get(pk=user)
             kwargs["user_form"] = VolunteerChangeForm(instance=volunteer_profile)
+            if kwargs["volunteer"].badges is not None:
```
joeleu, line 54:
```suggestion
if kwargs["volunteer"].badges:
```

You can rely on the truthy nature of the type instead of using the negative boolean check.

voluncheer/profiles/views/profile.py
```diff
@@ -101,26 +104,76 @@ def password_reset_request(request):
...
+        profile = get_object_or_404(Organization, pk=request.user)
         form = OrganizationChangeForm(request.POST, request.FILES, instance=profile)
     else:
         raise ValueError("profile_update: user must either a volunteer or an organizaiton.")
     form.save()
     return redirect("profile")
 
 
+"""
+This part is for Volunteer specified features.
```
joeleu, line 124: For non-docstring comments use the # format: https://peps.python.org/pep-0008/#block-comments

voluncheer/profiles/views/profile.py
```diff
@@ -101,26 +104,76 @@ def password_reset_request(request):
...
     )
+
+
+"""
+This part is for Organization specified features.
+"""
+
+
+def confirm_attandance(request, opportunity_id):
+    confirm_attendees = request.GET.getlist("volunteer-attanded")
```
joeleu, line 149:
```suggestion
    confirm_attendees = request.GET.getlist("volunteer-attended")
```

voluncheer/templates/navbar.html
```diff
@@ -15,13 +15,13 @@
       </a>
     </li>
     <li>
-      <a href="messages">
+      <a class="not-implemented-yet" href="#">
```
joeleu, line 18: Let's just remove these, adding them back once useful is trivial.

voluncheer/templates/profiles/profile.html
```diff
@@ -26,7 +25,7 @@
                 <hr align="center">
             </div>
             <div class="url">
-                <a href="#settings">IMPACT</a>
+                <a class="not-implemented-yet" href="#">IMPACT</a>
```
joeleu, line 28: Same here, better to remove now. We can re-add them later.

joeleu (changes requested):

voluncheer/opportunityboard/models.py
JinweiZhu: Right. I totally agree with Joe on this.

JinweiZhu (commented):
Good work. Just som minor tweak needed.

voluncheer/profiles/models.py
dbleek, line 1: Just FYI, I changed the field type for badges in my PR. I removed badges from the forms altogether because it should be added by the app, not directly during user signup/update, so the bug is no longer an issue.

voluncheer/profiles/views/profile.py
```diff
@@ -47,16 +49,17 @@ def get_context_data(self, **kwargs):
...
             kwargs["user_form"] = VolunteerChangeForm(instance=volunteer_profile)
+            if kwargs["volunteer"].badges is not None:
+                badge_urls = []
+                badge_list = kwargs["volunteer"].badges.split(",")
+                for badge in badge_list:
+                    try:
+                        badge_urls.append(Volunteer.BADGES[badge])
+                    except KeyError:
+                        pass
+                kwargs["badge_urls"] = badge_urls
```
dbleek, line 62: I rewrote the logic for passing badges to the template here, so most of it will be rewritten when we merge our branches.

voluncheer/profiles/urls.py
```diff
@@ -5,6 +5,7 @@
 from profiles.views.home import home
 from profiles.views.organizations import OrganizationSignUpView
 from profiles.views.profile import ProfileView
+from profiles.views.profile import confirm_attandance
```
dbleek, line 8: Slight typo, the correct spelling is "attendance."

dbleek (commented):
Seems like we will need to change the approach to archiving the opportunity, which will then change some other parts of the code. What about adding some logic in the view where after the date has passed, the opportunity appears in an "past opportunities" section? Or you could add a boolean flag for `is_archived` to the opportunity model.

voluncheer/opportunityboard/models.py
EdenWuyifan: I've tried to inherit the opportunity model, however, when creating an Archive object it will also create an Opportunity(like Volunteer to User). Hence it will create duplicate object to opportunityboard and profile. I think that will cause more issues.

EdenWuyifan:
> is_archived
Yeah, that's one solution to this

voluncheer/opportunityboard/models.py
joeleu: OK, using a set of fields to identify archived opportunities would have been my preference anyway.

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -83,6 +76,12 @@ def deregister_volunteer(request, opportunity_id):
     opportunity = Opportunity.objects.get(pk=opportunity_id)
     if volunteer in opportunity.volunteers.all():
         opportunity.volunteers.remove(volunteer)
+        # delete archived objects
```
joeleu, line 79: I don't understand what we are trying to do here, would you mind explaining it?

voluncheer/profiles/urls.py
```diff
@@ -26,4 +27,5 @@
         name="volunteer_signup",
     ),
     path("savedevents/", saved_events, name="saved_events"),
+    path("attandance/<int:opportunity_id>", confirm_attendance, name="attandance"),
```
Hgersten-hash, line 30: switch url to attendance

voluncheer/opportunityboard/views/search.py
```diff
@@ -60,7 +60,8 @@ def search(self):
         subsubcategory(default: None)
         Output: Opportunity list
         """
-        filtered_opportunity = Opportunity.objects.all()
+        filtered_opportunity = Opportunity.objects.filter(is_archive=False)
```
joeleu, line 63: Please update the docstring to reflect that this only searches un-archived opportunities.

voluncheer/profiles/tests/test_views.py
```diff
@@ -74,3 +78,39 @@ def test_signup_form_exists_at_desired_location(self):
...
+    def test_confirm_attendance(self):
+        """Test the confirm attendance feature.
+        The request here is for mimicing the get request of select a volunteer and submit.
+        The first confirm_attendance call won't work since volunteer is not signed up.
+        The second confirm_attendance call will add the volunteer to attended_volunteers.
+        The third confirm_attendance call will remove the volunteer from attended_volunteers.
+        """
+        rf = RequestFactory()
+        get_request = rf.get(
+            f"/attandance/{self.opp.pk}",
```
Hgersten-hash, line 98: should be f"/attendance/..."

voluncheer/profiles/tests/test_views.py
```diff
@@ -74,3 +78,39 @@ def test_signup_form_exists_at_desired_location(self):
         """Tests accessing home by URL"""
         response = self.client.get("")
         self.assertEqual(response.status_code, 200)
+
+
+class OrganizationProfileTest(TestCase):
+    """Test cases for Organization profile based functions"""
+
+    def setUp(self):
```
joeleu, line 86: Since you aren't extending this function it can be omitted.

voluncheer/templates/profiles/profile.html
```diff
@@ -74,7 +73,28 @@ <h2>POSTED OPPORTUNITIES</h2>
                         <p class="vol-position-disciption-h2">Date: {{ opportunity.date }}</p>
                         <p class="vol-position-disciption-h2">End: {{ opportunity.end }}</p>
                         <p class="vol-position-disciption-h2">{{ opportunity.description }}</p>
-                        <div class="url"><a href="#" class="readmore">Connect</a></div>
+                        <form method="get" id="opportunities-detail-{{opportunity.pk}}-form" class="opportunities-detail-form" action="{% url 'attandance' opportunity.pk %}">
+                            <div id="volunteer-attandence">
+                                <p class="vol-position-disciption-h1">Volunteer Attandance:</p>
```
Hgersten-hash, line 78: Volunteer Attendance

voluncheer/profiles/views/profile.py
```diff
@@ -97,26 +100,81 @@ def password_reset_request(request):
...
+    """
+    confirm_attendees = request.GET.getlist("volunteer-attended")
+    organization = get_object_or_404(Organization, pk=request.user.pk)
+    opportunity = get_object_or_404(Opportunity, pk=opportunity_id)
+    for attendee_pk in confirm_attendees:
+        volunteer = Volunteer.objects.get(pk=attendee_pk)
+        if volunteer not in opportunity.volunteers.all():
+            continue
+        if volunteer in opportunity.attended_volunteers.all():
+            opportunity.attended_volunteers.remove(volunteer)
```
joeleu, line 152: Why have they no longer attended the opportunity here?

voluncheer/profiles/views/profile.py
```diff
@@ -97,26 +100,81 @@ def password_reset_request(request):
...
+            Opportunity.objects.filter(
+                is_archive=True, volunteer_archive=volunteer, title=opportunity.title
+            ).delete()
+            volunteer.hours_volunteered -= opportunity.duration
```
joeleu, line 156: The volunteers hours are decreasing after participating in an opportunity?

voluncheer/profiles/views/profile.py
```diff
@@ -97,26 +100,81 @@ def password_reset_request(request):
...
+            volunteer.save()
+        else:
+            opportunity.attended_volunteers.add(volunteer)
+            print(f"{volunteer.hours_volunteered} + {opportunity.duration}")
```
joeleu, line 160: Leftover print debugging I assume?

voluncheer/profiles/forms/organizations.py
```diff
@@ -37,8 +37,6 @@ class OrganizationChangeForm(UserChangeForm):
 
     password = None
 
-    photo = forms.ImageField(required=False)
```
joeleu, line 40: Why are these being removed?

voluncheer/profiles/forms/volunteers.py
```diff
@@ -51,7 +51,6 @@ def save(self, commit=True):
 
 class VolunteerChangeForm(UserChangeForm):
     password = None
-    photo = forms.ImageField(required=False)
```
joeleu, line 54: Here too

voluncheer/profiles/tests/test_views.py
```diff
@@ -74,3 +78,39 @@ def test_signup_form_exists_at_desired_location(self):
...
+        """
+        rf = RequestFactory()
+        get_request = rf.get(
+            f"/attandance/{self.opp.pk}",
+            {"volunteer-attended": [str(self.vol.pk)]},
+        )
+        get_request.user = self.org
+
+        confirm_attendance(get_request, self.opp.pk)
+        self.assertEqual(self.opp.attended_volunteers.count(), 0)
```
joeleu, line 104: I don't understand why we are removing volunteers from an opportunity once they have attended it. Organizations are going to want the ability to view the history of their events and this seems counter productive to that. Or do I totally misunderstand what is going on here?

joeleu (changes requested):

voluncheer/profiles/models.py
```diff
@@ -182,7 +182,7 @@ class Volunteer(models.Model):
     badges = models.ManyToManyField(Badge, related_name="badges", blank=True)
     photo = models.ImageField(upload_to=_profile_photo_path, blank=True, null=True)
     description = models.TextField(help_text="Introduce yourself here.", default="")
-    hours_volunteered = models.DurationField(default=timedelta(days=0))
+    hours_volunteered = models.DurationField(default=timedelta(days=0), null=True)
```
Hgersten-hash, line 185: Why are we allowing null values?

PR #182, opened 2023-04-09
Add commands for loading categories and creating super users.

joeleu:


voluncheer/opportunityboard/management/commands/load_categories.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class Command(BaseCommand):
+    help = "Loads missing categories for the opportunity board."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--category_data",
+            default="opportunityboard/fixtures/category_data.json",
+            help="The email address of the super user to create.",
```
dbleek, line 24: I think this made it over from the su command.

dbleek (approved):
Nice--I see some things here I can add to/change about my badge data command/test. There's one small error in the categories command, but I'm approving preemptively since it's minor and everything else looks good.

voluncheer/opportunityboard/management/commands/load_categories.py
joeleu: Hah, failed copy/pasta. Thanks for noticing and mentioning. Fixed.

PR #196, opened 2023-04-14
added autocomplete features

suryasdas:


voluncheer/chatroom/migrations/0001_initial.py
```diff
@@ -1,12 +1,11 @@
 # Generated by Django 3.2.8 on 2023-03-30 16:52
 
+import django.db.models.deletion
```
JinweiZhu, line 3: Modifying existing migration files will break the RDS on AWS. I am not sure if reformatting them would but to be safe, make sure you open the INET project of VScode, so that way when you run black or isort, it will look at pyproject.toml file which asks it to ignore the migration files.

JinweiZhu (changes requested):

voluncheer/chatroom/migrations/0001_initial.py
suryasdas: do I have to make any changes anywhere again before I update the code?

PR #211, opened 2023-04-16
Chatroom on aws

JinweiZhu:
My plan is as follows: Use docker container to host the chatroom locally.
On AWS EBS: According to the Procfile, on aws, daphne will serve asgi and gunicorn will serve wsgi.
We should merge this last because this might break dev.


joeleu:
> My plan is as follows: Use docker container to host the chatroom locally. On AWS EBS: According to the Procfile, on aws, daphne will serve asgi and gunicorn will serve wsgi. We should merge this last because this might break dev.

If you haven't been able to get this to work on your own instance maybe we shouldn't merge it at all. "We should merge this last because this might break dev." does not inspire confidence... If that is a real risk we should perform better testing prior to merging into develop.

JinweiZhu:
> > My plan is as follows: Use docker container to host the chatroom locally. On AWS EBS: According to the Procfile, on aws, daphne will serve asgi and gunicorn will serve wsgi. We should merge this last because this might break dev.
> 
> If you haven't been able to get this to work on your own instance maybe we shouldn't merge it at all. "We should merge this last because this might break dev." does not inspire confidence... If that is a real risk we should perform better testing prior to merging into develop.

We cannot really test daphne locally because I run into port issues. I suppose on aws the situation will be different.

voluncheer/voluncheer/settings.py
```diff
@@ -234,6 +232,7 @@
 if environment.is_local:
     STATIC_URL = "/static/"
     STATICFILES_DIRS = [os.path.join(BASE_DIR, "static")]
+    # STATIC_ROOT = os.path.join(BASE_DIR, "static")
```
joeleu, line 235: Please remove instead of comment.

voluncheer/voluncheer/settings.py
```diff
@@ -55,6 +58,7 @@
     "crispy_bootstrap5",
     "storages",
     "channels",
+    "supervisor",  # Not sure if supervisor is needed here.
```
joeleu, line 61: If you aren't sure we should probably remove it and only add it once we are convinced we need it. It's easier to add code than remove it.

voluncheer/voluncheer/asgi.py
```diff
@@ -8,16 +8,21 @@
...
 from channels.routing import ProtocolTypeRouter
 from channels.routing import URLRouter
+import django
 from django.core.asgi import get_asgi_application
 
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "voluncheer.settings")
+django.setup()
+
+django_asgi_app = get_asgi_application()
 from chatroom import routing
```
joeleu, line 21: Does this import _need_ to be below the above lines? Imports should be at the top of the file.

voluncheer/.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 100
 exclude = */migrations/*
-ignore = E203
+ignore = E203,E402
```
joeleu, line 4: We should not ignore this across the project. If you absolutely must use it, use it locally for single exceptions.

joeleu (changes requested):
Please try deploying this to your own AWS instance before we try this on the develop branch, that would give us a little more confidence to deploy this to our environments.

voluncheer/voluncheer/asgi.py
JinweiZhu: Yes, if I don't get_asgi_application first, then there is no chatroom.

voluncheer/voluncheer/settings.py
JinweiZhu: I have the config file for supervisor anyway. I think we should leave it there?

JinweiZhu:
> Please try deploying this to your own AWS instance before we try this on the develop branch, that would give us a little more confidence to deploy this to our environments.

I did that on my own EBS. Would you approve now so I can merge?

voluncheer/voluncheer/settings.py
joeleu: But what is it doing? Why is it there?

voluncheer/voluncheer/settings.py
```diff
@@ -97,26 +101,20 @@
...
-    CHANNEL_LAYERS = {
-        "default": {
-            "BACKEND": "django_redis.cache.RedisCache",
-            "LOCATION": REDIS_CHATROOM_PORT,
-            "OPTIONS": {
-                "CLIENT_CLASS": "django_redis.client.DefaultClient",
-            },
-        }
-    }
+    REDIS_CHATROOM_PORT = os.getenv("REDIS_CHATROOM_PORT")
```
joeleu, line 105:
```suggestion
    REDIS_CHATROOM_PORT = os.getenv("REDIS_CHATROOM_PORT", "127.0.0.1")
```

Then you can remove the branching logic.

joeleu (dismissed):
I don't really know what the config files do, can you point me to a reference you used to create these? I want to understand how you selected the various properties, etc.

EdenWuyifan (approved):

PR #212, opened 2023-04-16
Archive opportunities

Hgersten-hash:
This pr creates a django command that will archive opportunities if it's end time and date have passed. It also creates a cron job and adds it to the django.config file. First it removes any old archive cron job if it exists and then it adds the archive_opportunities, followed by a command to set the permissions for this cron job. For now, the cron job is set to run at midnight.


voluncheer/remove-local-db-and-rebuild-db.sh
JinweiZhu, line 1: Why remove the content of this file? For example, the docker and load fixture commands are still useful.

Hgersten-hash: I don't know why it's saying I changed this file. I did not touch it

joeleu: This looks like the size changed but the contents didn't. Maybe you changed the permission (chmod +x for instance)? Did you maybe edit some metadata around the file?

Hgersten-hash: I did do chmod +x to run the script. This changes the file?

JinweiZhu: Right, it just changes permission no big deal.

JinweiZhu (approved):

voluncheer/opportunityboard/test_commands.py
```diff
@@ -38,3 +46,120 @@ def test_load_categories(self):
...
+
+        self.now = datetime.now(pytz.utc)
+        description = "Please help us support our community at this week's soup kitchen"
+
+        self.opportunity1 = Opportunity.objects.create(
+            organization=self.org,
+            category=Category.objects.filter(name=_CATEGORY).first(),
+            title="Test opportunity 1",
+            description=description,
+            date=timezone.now() - timedelta(days=1),
```
dbleek, line 72: I suggest using fixed dates for the datetimes. Although the logic is correct as is, using fixed dates will ensure that the tests run consistently.

For example:
```
today = datetime(year=2023, month=4, day=16, tzinfo=dt.timezone.utc)
...
self.opportunity1 = Opportunity.objects.create( ...
date = today - timedelta(days=1)
```

voluncheer/opportunityboard/management/commands/archive_opportunities.py
```diff
@@ -0,0 +1,31 @@
...
+from django.core.management.base import BaseCommand
+from django.db import transaction
+import pytz
+
+from opportunityboard.models import Opportunity
+
+
+class Command(BaseCommand):
+    help = "Archive opportunities that have ended"
+
```
dbleek, line 12: You could add an argument to run the command for a given day for testing purposes (see comment below). 

```
    def add_arguments(self, parser):
        parser.add_argument(
            "--datetime",
            default=datetime.now(pytz.utc),
            help="The date before which all opportunities are archived",
        )
```

You would then use` options["datetime"]` in place of `now `elsewhere in the command.

voluncheer/opportunityboard/test_commands.py
dbleek: That said, this would only work if you made the above changes to the command so you could call it for a specific date.

dbleek (approved):
Looks great. I have some suggestions for improving the command and tests to avoid using timezone.now(), but it's more theoretical than anything else. I have been told to never use timezone.now in tests and I think it's a good a rule to keep to. Going to approve regardless but please consider making the suggested changes.

voluncheer/opportunityboard/test_commands.py
Hgersten-hash: Out of curiosity, why would it be bad to use timezone for tests?

voluncheer/opportunityboard/management/commands/archive_opportunities.py
Hgersten-hash: This is an interesting idea! I will applied the changes. WYTD?

voluncheer/opportunityboard/test_commands.py
dbleek: Because it is non-deterministic and the test will never be exactly the same any time you run it. Say for example you used `datetime.now()` and asserted it was before a fixed date well in the future. Your test could be successful for years, but eventually when the fixed date passes, it would no longer pass. 

This isn't an issue in this case since you aren't mixing in fixed dates, but just an example to illustrate how the test is no longer locked in place when using `now()`.

dbleek:
Looks good to me

PR #217, opened 2023-04-17
Archive opportunities

Hgersten-hash:
Updated the test to use the freezegun lib. Also, set the cron job to run hourly to make sure it works.

### DIFF
--- a/voluncheer/.ebextensions/django.config
+++ b/voluncheer/.ebextensions/django.config
@@ -14,7 +14,7 @@ container_commands:
   05_remove_old_archive_cron:
     command: "rm -f /etc/cron.d/archive_opportunities"
   06_create_archive_cron:
-    command: "echo '0 0 * * * root source /var/app/venv/*/bin/activate && python3 manage.py archive_opportunities' > /etc/cron.d/archive_opportunities"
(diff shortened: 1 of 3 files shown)


voluncheer/.ebextensions/django.config
```diff
@@ -14,7 +14,7 @@ container_commands:
   05_remove_old_archive_cron:
     command: "rm -f /etc/cron.d/archive_opportunities"
   06_create_archive_cron:
-    command: "echo '0 0 * * * root source /var/app/venv/*/bin/activate && python3 manage.py archive_opportunities' > /etc/cron.d/archive_opportunities"
+    command: "echo '0 * * * * root source /var/app/venv/*/bin/activate && python3 manage.py archive_opportunities' > /etc/cron.d/archive_opportunities"
```
joeleu, line 17: How are you going to confirm this is running? Are you planning to create an opportunity you expect to be archived and then check it's status after an hour?

joeleu (approved):

voluncheer/.ebextensions/django.config
Hgersten-hash: Yes, that's what I was thinking. I actually just changed it to run every ten minutes so that I can check the results earlier

joeleu: SGTM

PR #218, opened 2023-04-17
fix the display problem after merging filter by distance

EdenWuyifan:


voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -93,6 +93,8 @@ def deregister_volunteer(request, opportunity_id):
 def organization_view(request, userid):
     organization = Organization.objects.get(pk=userid)
     user = get_object_or_404(User, pk=request.user.pk)
+    if user.is_organization:
```
JinweiZhu, line 96: Why do we not allow an organization to see the list of opportunities?

EdenWuyifan: Nonono, this is for banning organization viewing volunteer-viewed organization page

Hgersten-hash (approved):

PR #230, opened 2023-04-18
add past opportunity section

EdenWuyifan:
image

image


voluncheer/templates/profiles/profile.html
```diff
@@ -137,6 +137,28 @@ <h2>SAVED OPPORTUNITIES</h2>
...
                 {% endfor %}
         </div>
+
+        <h2>PAST OPPORTUNITIES</h2>
+        <div class="card" id="saved_opportunities">
+            {% csrf_token %}
+                {% for opportunity in opportunity_lists %}
+                    {% if opportunity.is_archived %}
+                    <article>
+                        <a href="{% url 'update_an_opportunity' opportunity.pk %}"><i id="opportunity-edit-{{opportunity.pk}}" class="fa fa-pen fa-xs edit"></i></a>
```
dbleek, line 147: Should we allow users to edit past opportunities? I can think of a reason or two to say yes, but it might be safer to not allow changes after it has passed.

voluncheer/templates/profiles/profile.html
```diff
@@ -137,6 +137,28 @@ <h2>SAVED OPPORTUNITIES</h2>
                     {% endif %}
                 {% endfor %}
         </div>
+
+        <h2>PAST OPPORTUNITIES</h2>
+        <div class="card" id="saved_opportunities">
+            {% csrf_token %}
```
dbleek, line 143: It looks like we only need to use this in POST forms, there isn't any thing here that we would need to protect, unless I'm missing something. 

https://docs.djangoproject.com/en/4.2/howto/csrf/

voluncheer/templates/voluncheer/opportunityboard.html
dbleek, line 1: This looks good, but we should double check with the Hanna and the team if we want this.

dbleek (changes requested):

voluncheer/templates/profiles/profile.html
Hgersten-hash: I don't think they should be allowed to edit those. Once it has passed it's over; no reason to edit it. The only two things I can see an Organization wanting to do is copying one, or removing it from public site.

dbleek: @EdenWuyifan could you please let us know what the status is on this?

EdenWuyifan: Fixed!

voluncheer/templates/profiles/profile.html
EdenWuyifan: Deleted

voluncheer/templates/voluncheer/opportunityboard.html
Hgersten-hash: I like having a display of the badges, but I don't think the best place for it is right by the opportunity card. Maybe instead we can have it on the side - next to the board?

EdenWuyifan: That make sense. I will fix this in next sprint

Hgersten-hash (approved):

dbleek (approved):

PR #231, opened 2023-04-19
Bugfix dob validation

dbleek:
This PR addresses an issue raised during the testing party where users can enter a date of birth in the future when signing up or updating their profile. It also adds a date picker to the date of birth field in both the creation and update forms.


EdenWuyifan (commented):
https://www.geeksforgeeks.org/html-dom-input-datetimelocal-max-property/

Referring to this, you can directly add "min="2019-02-18T10:30:55" max=....." to the input field, or edit the min max property using Javascript after locating the element

dbleek:
Interesting, yeah I saw I something like that when looking into how using flatpickr. I guess if we had a lot of date fields in one template, it would be easier to do write a little bit of JS in the template to find and change the type of all the fields. Is there an advantage to doing it that way here, where we only need to change one date for each template?

EdenWuyifan (approved):

PR #234, opened 2023-04-20
Sign up bug fix

Hgersten-hash:
This pr fixes the bug that users were created even if the email could not be sent. It also adds a message template to handle messages sent to the user.


voluncheer/profiles/views/organizations.py
```diff
@@ -18,10 +19,19 @@ def get_context_data(self, **kwargs):
...
     def form_valid(self, form):
         """Saves the new user and logs them in."""
         user = form.save()
         user.is_active = False
         user.save()
-        activateEmail(self.request, user, form.cleaned_data.get("email"))
+        try:
+            activateEmail(self.request, user, form.cleaned_data.get("email"))
+        except Exception:
+            # Delete the user if an error occurs
```
EdenWuyifan, line 31: Remove the commented lines for final merging

voluncheer/profiles/views/volunteers.py
```diff
@@ -20,10 +20,19 @@ def get_context_data(self, **kwargs):
...
         """Saves the new user and logs them in."""
         user = form.save()
         user.is_active = False
         user.save()
-        activateEmail(self.request, user, form.cleaned_data.get("email"))
+        try:
+            activateEmail(self.request, user, form.cleaned_data.get("email"))
+        except Exception:
+            # Delete the user if an error occurs
+            print("here!")
```
EdenWuyifan, line 33: Here!

voluncheer/profiles/tests/test_email_verification.py
```diff
@@ -48,3 +55,17 @@ def test_email_sent(self):
             reverse("activate", args=[uidb64, token]),
             mail.outbox[0].body,
         )
+
+    @patch.object(activate_email, "activateEmail")
+    def test_organization_signup_view_email_error(self, mock_email):
+        mock_email.exists.side_effect = Exception("Server Error")
+        self.form_response = OrganizationSignUpView.form_valid(self.request, form=self.form1)
+        self.assertEqual(self.form_response.status_code, 302)
```
EdenWuyifan, line 63: Here should also add 200 OK to a GOOD status code

voluncheer/profiles/tests/test_email_verification.py
```diff
@@ -48,3 +55,17 @@ def test_email_sent(self):
...
+        self.assertFalse(User.objects.filter(email="test_org@testing.org").exists())
+
+    @patch.object(activate_email, "activateEmail")
+    def test_volunteer_signup_view_email_error(self, mock_email):
+        mock_email.exists.side_effect = Exception("Server Error")
+        self.form_response = VolunteerSignUpView.form_valid(self.request, form=self.form1)
+        self.assertEqual(self.form_response.status_code, 302)
```
EdenWuyifan, line 70: Here also

EdenWuyifan (changes requested):
Some minor tweaks. LGTM!

voluncheer/profiles/tests/test_email_verification.py
Hgersten-hash: I am not sure what you mean. The form_valid function will redirect to signup if the email could not be sent, or it will redirect to login if the email was correct. Can you explain what you mean by 200?

voluncheer/profiles/tests/test_email_verification.py
Hgersten-hash: same question as above

voluncheer/profiles/tests/test_email_verification.py
EdenWuyifan: 200 is the normal HTTP status code when the page is correctly loaded, 302 is technically not a good status however we should omit that since we constantly using redirect. But still adding 200 to the assert is more firmly.
Referring to this post: https://blog.short.io/status-codes/

You can use this line instead:
self.assertIn(response.status_code, [200, 302])

EdenWuyifan (approved):

PR #238, opened 2023-04-23
Do not show opportunities before now on the opportunity board.

joeleu:
This PR stops opportunities that start now or in the past from displaying on the opportunity board to prevent users from seeing opportunities they effectively cannot participate in given the arrow of time.

### DIFF
--- a/voluncheer/opportunityboard/tests_views.py
+++ b/voluncheer/opportunityboard/tests_views.py
@@ -1,8 +1,9 @@
-import datetime as dt
+import datetime
 
 from django.contrib.auth import get_user_model
 from django.test.client import RequestFactory
 from django.urls import reverse
+import freezegun
 
 from opportunityboard.models import Category
 from opportunityboard.unittest_setup import TestCase
@@ -49,6 +50,12 @@ class OpportunityboardTestCase(TestCase):
         response = opportunityboard(get_request, 1)
         self.assertEqual(response.status_code, 200)
 
+    def test_search_excludes_opportunities_before_now(self):
+        """Tests that opportunities before now are not returned."""
+        with freezegun.freeze_time(self.date + datetime.timedelta(days=1)):
+            filters = Filter()
+            self.assertEqual(filters.search().count(), 0)
+
     def test_search_by_categories(self):
         """Tests filter search by category/subcategory/subsubcategory functions"""
         filters = Filter()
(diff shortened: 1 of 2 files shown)


voluncheer/opportunityboard/tests_views.py
```diff
@@ -49,6 +50,12 @@ def test_search(self):
         response = opportunityboard(get_request, 1)
         self.assertEqual(response.status_code, 200)
 
+    def test_search_excludes_opportunities_before_now(self):
+        """Tests that opportunities before now are not returned."""
+        with freezegun.freeze_time(self.date + datetime.timedelta(days=1)):
```
dbleek, line 55: 😮

dbleek: Also, I can't figure out where self.date is coming from. Am I missing something?

dbleek: Does using `with` ensure that the freezegun is stopped after the test?

voluncheer/opportunityboard/views/search.py
```diff
@@ -87,6 +88,14 @@ def search(self):
         return filtered_opportunity
 
 
+def filter_opportunities_before_now(queryset):
+    now = datetime.datetime.now().timestamp()
```
dbleek, line 92: Why do you use a timestamp here as opposed to just using `now` in the comparison?

dbleek (commented):
All looks good, leaving open mostly for discussion/edification purposes. 

Also it seems like you don't like abbreviating `datetime`. Is that considered bad practice?

voluncheer/opportunityboard/tests_views.py
joeleu: `self.date` is coming from the `setUp` method on the base `TestCase` class.

Yes, exactly. Using `freezegun` as a context manager (with the `with` statement) boxes the functionality to the code indented below it (because Python scope is based on whitespace).

voluncheer/opportunityboard/views/search.py
joeleu: Because I don't know the timezone for each opportunity at this point. This grabs a naive timestamp (UTC) and then for each opportunity we convert it to the timezone of the opportunity during comparison.

joeleu:
> All looks good, leaving open mostly for discussion/edification purposes.
> 
> Also it seems like you don't like abbreviating `datetime`. Is that considered bad practice?

I prefer not using abbreviations whenever possible. When an abbreviation is used I have to find where it was declared to be sure I know what I'm reading whereas if the exact module name is used I can implicitly assume it is the standard library. Overall abbreviations require the reader to find the declaration even when reading an isolated bit of code. Plus unless they are all kept exactly the same it makes refactoring more difficult.

Here is Google's style guide explaining a bit about imports for a bit of useful context: https://google.github.io/styleguide/pyguide.html#22-imports

voluncheer/opportunityboard/tests_views.py
dbleek: Ah sorry I missed my own import statement :)

Thanks!

voluncheer/opportunityboard/views/search.py
dbleek: Interesting, thank you

dbleek (approved):
Thanks again for the insight.

PR #239, opened 2023-04-23
Volunteers Can View Past Opportunities

dbleek:
This PR allows volunteers to view opportunities under the My Voluncheer Footprints section when:
- the opportunity is archived
- the volunteer is marked as having attended the opportunity


joeleu (approved):
Excellent work!

PR #249, opened 2023-04-25
Merge Supervisor Script into Develop

dbleek:
This PR would confirm if the supervisor script, which is the only difference in the codebase between main and develop, is causing any problems with chatroom. We can revert develop to the previous commit once testing is complete.


Hgersten-hash:
Wayne and I merged the supervisor script into main_v2 yesterday to try this.

Hgersten-hash:
Well, originally, the supervisor was in develop_v2 and it was working. Then wayne removed it to see if the env would still work and it did. So, when he created main_v2 he didn't include the supervisor which is why we tried to merge the supervisor back into main_v2 to see if it would help

PR #250, opened 2023-04-25
Archive opportunities cron job

Hgersten-hash:
This pr adds a cron job to archive opportunities. For now, it is scheduled to run every 10 minutes for debugging purposes.


voluncheer/.ebextensions/03_archive_cron.config
joeleu, line 1: I don't really understand this but it seems fine to me.

voluncheer/static/fonts/ZenKakuGothicNew-Black.ttf
joeleu, line 1: Did you mean to remove these fonts? If yes, the removal should be moved to a different PR in case we need to revert either of the changes then unrelated changes won't also be reverted.

voluncheer/.ebextensions/03_archive_cron.config
Hgersten-hash: It creates two files. One is the script containing the instructions to run the django command, and the other creates the cron job. The command below removes any old cron versions of the cron job.

Hgersten-hash: Found tutorial from here and modified paths for our app:
[https://repost.aws/knowledge-center/cron-job-elastic-beanstalk](url)

PR #252, opened 2023-04-27
Add daphne logs

dbleek:
The PR should allow us to view the logs from Daphne.


Hgersten-hash (changes requested):
You might need to change the numbers in the file names as there is already a 03 config file named archive_chron

Hgersten-hash (approved):

PR #258, opened 2023-04-28
Feature eden recurring

EdenWuyifan:
- Create a series of opportunities upon posting a recurring opportunity. Max to 12 of them once. 
- Add a recurring opportunity section, wrapping the series of recurring opportunities(siblings). Allow users to expand and edit each one of them.
image
image
image
- Add a Delete Recurrences option on edit an opportunity page, allowing users to delete all siblings at once.
image


joeleu:
Can you provide a description of what this PR is meant to do? Or at the very least attach the issue for the feature you are implementing?

voluncheer/opportunityboard/models.py
```diff
@@ -119,3 +120,7 @@ def duration(self):
         duration = end_datetime - start_datetime
         duration_seconds = dt.timedelta(seconds=duration.seconds)  # remove date
         return duration_seconds
+
+    def delete_recurrences(self):
```
Hgersten-hash, line 124: Would this delete any siblings that have already been archived?

voluncheer/opportunityboard/views/opportunityboard.py
```diff
@@ -98,11 +98,24 @@ def organization_view(request, userid):
     if user.is_organization:
         return redirect("home")
     volunteer = get_object_or_404(Volunteer, pk=user)
-    opportunity_lists = organization.opportunity_set.all()
+    recurring_posted_opportunity_lists = organization.opportunity_set.filter(
```
Hgersten-hash, line 101: I think you should keep the original name "opportunity list" so that it is still clear what this list is for.

voluncheer/opportunityboard/views/postanopportunity.py
```diff
@@ -55,6 +55,8 @@ def update_an_opportunity(request, opportunity_id):
         if form.is_valid():
             if "delete" in request.POST:
                 form.delete(opportunity_id)
+            elif "delete_recurrences" in request.POST:
+                form.delete_recurrences(opportunity_id)
```
Hgersten-hash, line 59: again, would this delete archived opportunities that are siblings?

Hgersten-hash (changes requested):
LGTM. The one thing I am concerned about is deleting archived opportunities

voluncheer/opportunityboard/views/opportunityboard.py
dbleek: I agree!

voluncheer/profiles/tests/test_views.py
```diff
@@ -263,3 +265,30 @@ def test_volunteer_can_see_past_opportunities(self):
             self.opp.attended_volunteers.remove(self.vol)
             response = self.client.get(reverse("saved_events"))
             self.assertNotIn(self.opp, response.context["opportunity_attended"])
+
+    def test_delete_recurring_opportunities(self):
+        """test deleting the recurring opportunities.
+        first add a serie of opportunities, first delete one, then delete all siblings
```
dbleek, line 271: small typo, "serie" should be "series".

voluncheer/static/js/main.js
```diff
@@ -27,4 +27,17 @@ function ShowDetail(element) {
   } else {
     form.style.display = "none";
   }
+};
+
+function ShowSibling(element) {
+  var id = element.id;
+  var form = document.getElementById(id+"-siblings");
+  console.log(form.style.display);
```
dbleek, line 35: we can remove this` console.log`, which I assume was for debugging

voluncheer/templates/opportunityboard/postanopportunity.html
```diff
@@ -112,6 +112,23 @@ <h2>OPPORTUNITY DETAILS</h2>
                     });
                   });
                 </script>
+                <script>
```
dbleek, line 115: Can we just use a timedelta instead? Also, what if you made a separate .js file  to contain this for readability (see Wayne's room.js in the chatroom template).

dbleek (changes requested):
Looks good so far, some readability issue and also curious about Hanna's comment regarding not deleting archived opportunities--we shouldn't delete opportunities that are part of a recurrence and already passed because that would affect a volunteer's history.

voluncheer/opportunityboard/models.py
EdenWuyifan: Good point. I'll add a guard here

voluncheer/opportunityboard/views/postanopportunity.py
EdenWuyifan: This won't misdelete after the guard is added into form.delete_recurrences

voluncheer/static/js/main.js
EdenWuyifan: THanks for pointing out!!

voluncheer/templates/opportunityboard/postanopportunity.html
EdenWuyifan: There's no build-in time delta in javascript. And sure I can move all script to a separate file. Thanks!

Hgersten-hash:
Just checked it out locally and it looks amazing! One thing to change would be to have the opportunities appear in chronological order.

Hgersten-hash (approved):

dbleek (approved):

PR #259, opened 2023-04-29
Add badge progression bar on the volunteer's profile.

joeleu:
For example: 
Screenshot 2023-04-29 at 17 34 00

### DIFF
--- a/voluncheer/profiles/tests/test_views.py
+++ b/voluncheer/profiles/tests/test_views.py
@@ -1,13 +1,14 @@
+import datetime
+
 from django.contrib.auth import get_user_model
 from django.test.client import RequestFactory
 from django.urls import reverse
 
-from opportunityboard import unittest_setup  # noqa:F401
 from opportunityboard.unittest_setup import TestCase
 from profiles.models import GalleryPost
 from profiles.models import UserType
+from profiles.views import profile
 from profiles.views.gallery import create_post
-from profiles.views.profile import confirm_attendance
 
 
 class SignupViewTest(TestCase):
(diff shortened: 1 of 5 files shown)


voluncheer/templates/base.html
```diff
@@ -7,7 +7,7 @@
     <title>Voluncheer</title>
     <link rel="stylesheet" href="{% static 'css/app.css' %}">
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.1/css/all.min.css">
-    <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>
+    <script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/3.6.4/jquery.min.js"></script>
```
EdenWuyifan, line 10: Why are we changing jquery version here?

EdenWuyifan (commented):
LGTM. Please test the change of jquery version is not affecting other functions

voluncheer/templates/base.html
joeleu: I changed it during debugging to see if it would help. Changing it did resolve a persistent error I see when looking at the Chrome debugging console (something about tooltip) but I don't believe the newer version is necessary to support these changes so I'll revet it.

voluncheer/profiles/tests/test_views.py
```diff
@@ -263,3 +264,23 @@ def test_volunteer_can_see_past_opportunities(self):
             self.opp.attended_volunteers.remove(self.vol)
             response = self.client.get(reverse("saved_events"))
             self.assertNotIn(self.opp, response.context["opportunity_attended"])
+
+    def test_badge_progression(self):
+        with self.subTest("hours_until_first_badge"):
+            self.vol.hours_volunteered = datetime.timedelta(hours=10)
+            progress, hours_remaining, _ = profile.badge_progression(self.vol)
+            self.assertEqual(progress, 20)
```
dbleek, line 272: It was unclear that progress was percentage until reading the rest of the code, especially back to back with hours_remaining. Perhaps a docstring could elaborate that the test is for returning correct hours remaining until next badge and progress percentage?

dbleek (changes requested):

voluncheer/profiles/tests/test_views.py
joeleu: I've added a bit more detail to the implementation. WDYT?

dbleek (approved):
Looks good!

PR #261, opened 2023-04-30
Add a message when the chatroom is empty and don't redirect.

joeleu:
Is it a little hacky? Yes, but it works :)

Ideally this validation would be done on the backend but since it is a get request and not a form I'm not entirely sure how to do that so here is a temporary solution.

This prevents the user from attempting to join a chatroom with an empty string for the name.

Screenshot 2023-04-30 at 12 15 37


Hgersten-hash (approved):
LGTM! At this point, hacky works just fine

PR #267, opened 2023-05-01
Archive opportunities

Hgersten-hash:
This pr updates the timezone for archive_opportunities command to the voluncheer settings timezone


voluncheer/.ebextensions/03_archive_cron.config
```diff
@@ -8,6 +8,7 @@ files:
             source /var/app/venv/*/bin/activate
             cd /var/app/current
             export $(cat /opt/elasticbeanstalk/deployment/env | xargs)
+            export $(cat /opt/elasticbeanstalk/deployment/env | xargs)
```
joeleu, line 11: This is here twice, seems like an oversight.

joeleu (approved):

voluncheer/.ebextensions/03_archive_cron.config
Hgersten-hash: Good catch!

PR #272, opened 2023-05-03
Store chat history

JinweiZhu:
I've tested this and it looks good. Let's merge it and then see what will happen?


voluncheer/chatroom/test_views.py
```diff
@@ -1,27 +1,28 @@
...
             email="test@voluncheer.com",
             password="secret",
             type=UserType.VOLUNTEER,
         )
         self.user.save()
         self.client.login(email="test@voluncheer.com", password="secret")
-
-    def tearDown(self):
-        self.user.delete()
+        self.vol2 = Volunteer.objects.create(
```
JinweiZhu, line 19: So where exactly do you use vol2?

voluncheer/chatroom/views.py
```diff
@@ -20,11 +23,22 @@ def chat_homepage_view(request):
...
+    if user.is_organization:
+        organization = Organization.objects.get(pk=user)
+    elif user.is_volunteer:
+        volunteer = Volunteer.objects.get(pk=user)
     return render(
         request,
         "chatroom/room.html",
         {
+            "message": message,
             "rooms": Room.objects.all(),
```
JinweiZhu, line 38: Why do we need to pass other rooms in as well?

voluncheer/chatroom/consumers.py
```diff
@@ -33,16 +58,22 @@ def disconnect(self, close_code):
...
+        timestamp = test_data_json["timestamp"]
+        photo = test_data_json["photo"]
+        self.save_message(user, room, content, timestamp)
         # send chat message event to the room
         async_to_sync(self.channel_layer.group_send)(
             self.room_group_name,
             {
                 "type": "chat_message",
-                "message": message,
+                "message": content,
```
JinweiZhu, line 75: Content is a field of message. Why are we assigning it to "message"? is it called message in html?

voluncheer/chatroom/consumers.py
```diff
@@ -13,6 +17,27 @@ def __init__(self, *args, **kwargs):
         self.room_group_name = None
         self.room = None
 
+    def save_message(self, user, room, content, timestamp):
+        print(user, room, content, timestamp)
```
joeleu, line 21: This print statement seems superfluous, it won't be viewed by anyone.

voluncheer/chatroom/consumers.py
```diff
@@ -33,16 +58,22 @@ def disconnect(self, close_code):
             self.channel_name,
         )
 
-    def receive(self, text_data=None, bytes_data=None):
-        text_data_json = json.loads(text_data)
-        message = text_data_json["message"]
-
+    def receive(self, text_data, bytes_data=None):
+        test_data_json = json.loads(text_data)
+        print(test_data_json)
```
joeleu, line 63: Same here.

voluncheer/static/admin/js/room.js
```diff
@@ -37,14 +40,32 @@ chatMessageInput.onkeyup = function (e) {
...
+        "room": roomName,
+        "timestamp": date.toGMTString(),
+        "photo": photo,
+    });
+    chatSocket.send(msg);
+
     chatMessageInput.value = "";
-};
+}
+// clear the 'chatMessageInput' and forward the message
```
joeleu, line 63: These commented out lines should be removed.

joeleu (dismissed):

voluncheer/chatroom/test_views.py
JinweiZhu: I see this is to resolve the bug yesterday right?

voluncheer/chatroom/consumers.py
JinweiZhu: Message class has fields: user, room, content, timestamp, volunteer, organization. So I am a bit confused why we are assigning content to message and calling message.content in room.html

voluncheer/chatroom/test_views.py
EdenWuyifan: Yeah. It's not a good idea to add photo attribute to unittest setup. So I add this back.

PR #274, opened 2023-05-04
Bugfix markers

dbleek:
This PR fixes the opportunity markers issue on the map by checking if `longitude `and `latitude `exist before trying to create a marker.


voluncheer/map/views.py
```diff
@@ -27,16 +27,17 @@ def map(request):
...
-                {
-                    "latitude": opportunity.latitude,
-                    "longitude": opportunity.longitude,
-                    "title": opportunity.title,
-                    "address": opportunity.address_1,
-                    "type": opportunity.category.name,
-                    "name": opportunity.organization.name,
-                }
-            )
+            if opportunity.latitude and opportunity.longitude:
```
joeleu, line 30: Instead of nesting even further let's move this logic out into a single function, something like

```
def should_show_opportunity(opp):
    return opp.is_published and not opp.is_archivhed and opp.latitude and opp.longitude
```

WDYT?

joeleu (changes requested):

joeleu (approved):

PR #275, opened 2023-05-05
add chat button, fixed roomname display, fix photo link bug…

EdenWuyifan:
... add chat history and other frontend features


voluncheer/templates/volunteer/vol_org_view.html
```diff
@@ -78,5 +88,15 @@ <h2>POSTED OPPORTUNITIES</h2>
         alert('You already registered this event!');
       }
     };
+    // redirect to '/room/<roomInput>/'
+    function toggleChat() {
```
JinweiZhu, line 92: why do we need this toggleChat() when we already have toggleChatOrg() and toggleChatVol()?

EdenWuyifan: Good point. I'll remove that

JinweiZhu (approved):

PR #278, opened 2023-05-06
remove unused search fields and add search by keyword, fixed search c…

EdenWuyifan:
…ount

- Search bar is now usable!!! Also the search result count is fixed.
image

- Work place is still on filters board but isn’t implemented - Eden
- Date range filter is not implemented - Eden


JinweiZhu (approved):
Good job. LGTM.

PR #279, opened 2023-05-06
Fix duration filter

dbleek:
This PR fixes a bug with the search by duration filter. The bug was caused by the `duration` property in the opportunity model, which created a new datetime object but didn't make it time zone aware.


JinweiZhu (approved):
LGTM

PR #287, opened 2023-05-07
Fix sorting

JinweiZhu:
I fixed the profile page UI display inconsistency locally when clicking on the side nav bar, and I fix the sorting as well.

### DIFF
--- a/voluncheer/templates/profiles/posttogallery.html
+++ b/voluncheer/templates/profiles/posttogallery.html
@@ -23,14 +23,17 @@
     </div>
 
     <div class="sidenav-url">
+        {% if user.pk is curr_user.pk %}
         <div class="url">
             <a href="{% url 'home' %}">PROFILE</a>
             <hr align="center">
         </div>
+        {% endif %}
         <div class="url">
             <a href="{% url 'saved_events' %}" >MY VOLUNCHEERS</a>
             <hr align="center">
         </div>
+        
         <div class="url">
             <a href="{% url 'create_post' %}" class="active">CREATE POST</a>
             <hr align="center">
--- a/voluncheer/templates/profiles/posttogallery.html
+++ b/voluncheer/templates/profiles/posttogallery.html
@@ -11,7 +11,7 @@
             <img src="{{volunteer.photo.url}}" alt="profile pic" width="100" height="100"/>
         {% endif %}
         <div class="name">
-            {{ volunteer.first_name }}, {{ volunteer.last_name }}
+            {{ volunteer.first_name }} {{ volunteer.last_name }}
         </div>
         <div class="email">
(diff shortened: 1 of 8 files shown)


voluncheer/templates/profiles/posttogallery.html
```diff
@@ -27,15 +27,11 @@
...
-        <div class="url">
-            <a class="not-implemented-yet" href="#">SETTINGS</a>
-            <hr align="center">
-        </div>
         <div class="url">
             <a href="{% url 'saved_events' %}" >MY VOLUNCHEERS</a>
             <hr align="center">
         </div>
-        {% if user.pk is not curr_user.pk %}
+        {% if user.pk is not volunteer.pk %}
```
EdenWuyifan: Why are you changing curr_user to volunteer? I think curr_user is for checking if one is viewing their own profile. No need to differentiate between volunteer and organization.

EdenWuyifan (changes requested):

voluncheer/templates/profiles/profile.html
```diff
@@ -287,27 +283,30 @@ <h2>PAST OPPORTUNITIES</h2>
...
-            <div class="url">
-                <a href="#" onclick="toggleChatVol()">CHAT</a>
-                <hr align="center">
-            </div>
+                <div class="url">
+                    <a href="#" onclick="toggleChatVol()">CHAT</a>
+                    <hr align="center">
+                </div>
             {% endif %}
+            {% if user.pk is curr_user.pk %}
```
EdenWuyifan, line 304: Create Post is for all users. Every user can post on others gallery. This also goes to viewing gallery

voluncheer/templates/profiles/profile.html
```diff
@@ -287,27 +283,30 @@ <h2>PAST OPPORTUNITIES</h2>
         </div>
 
         <div class="sidenav-url">
+            {% if user.pk is curr_user.pk %}
```
EdenWuyifan, line 286: PROFILE shouldn't be restrained from others

EdenWuyifan (changes requested):

voluncheer/templates/profiles/profile.html
JinweiZhu: If it's a different user, then they wouldn't be able to click other user's profile right? Right now, it redirects to the user's own profile when they click on other's profile.

voluncheer/templates/profiles/profile.html
JinweiZhu: Wait, but this just doesn't make sense to me.... How come others can determine the content in my gallery but now myself?

EdenWuyifan (approved):

